PR #98, opened 2015-08-23
trivial: spelling fixes

martingkelly:
These typos were found and reported by Lintian in the Debian packaging
for gmpy. This patch fixes them.  Although these are not super
important, they will make it easier to see real warnings/issues with
a cleaner Lintian page.

Here's the relevant Lintian page:
https://lintian.debian.org/full/martin@martingkelly.com.html


casevh:
Thanks for the patch and for keeping gmpy up-to-date in Debian.

martingkelly:
Thanks!

PR #107, opened 2016-04-21
Fix typo: denominator -> divisor

mulkieran:
Signed-off-by: mulhern <amulhern@redhat.com>


casevh:
Thanks for the patch.

Issue #123, opened 2016-10-19
Add support for arbitrary objects defining a __mpz__ method


jdemeyer:
It should be possible for other projects to define a conversion to `gmpy2` types. As is typically done in Python, this could be done by a `__mpz__` method. That is, the function `mpz` should do something like (assuming also #124):

``` python
def mpz(x):
    # Existing code
    if isinstance(x, known_types):
        return known_conversion_to_mpz(x)

    # New code
    try:
        m = x.__mpz__
    except AttributeError:
        pass
    else:
        return m()
    try:
        x = int(x)
    except TypeError:
        pass
    else:
        return known_conversion_to_mpz(x)

    # Existing code
    raise TypeError("mpz() requires numeric or string argument")
```

videlec:
See #134

videlec:
this could be closed now that #134 is merged...

videlec:
this could be closed...

casevh:
Closing. Changes merged in #134.

PR #134, opened 2017-05-17
check for __mpz__ and __mpq__ in New

videlec:
This would allow any object to implement their own conversion to gmpy2. The branch just make `gmpy2.mpz(a)` so that it calls `a.__mpz__()` when the method exists (and similar for `mpq`). With the branch applied one can do

    >>> class A:
    ...    def __mpz__(self):
    ...        import gmpy2
    ...        return gmpy2.mpz('42')
    ...    def __mpq__(self):
    ...        import gmpy2
    ...        return gmpy2.mpq('13/12')
    >>> import gmpy2
    >>> gmpy2.mpq(A())
    mpq(13,12)
    >>> gmpy2.mpz(A())
    mpz(42)

(fixes issue #123)

### DIFF
--- a/src/gmpy2_cache.c
+++ b/src/gmpy2_cache.c
@@ -148,6 +148,10 @@ GMPy_MPZ_NewInit(PyTypeObject *type, PyObject *args, PyObject *keywds)
             return (PyObject*)GMPy_MPZ_From_PyStr(n, base, context);
         }
 
+        if (PyObject_HasAttrString(n, "__mpz__")) {
+             return (PyObject*) PyObject_CallMethod(n, "__mpz__", NULL);
+        }
+
         /* Try converting to integer. */
         temp = PyNumber_Long(n);
         if (temp) {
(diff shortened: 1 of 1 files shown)


videlec:
With the branch applied one can do

    >>> class A:
    ...    def __mpz__(self):
    ...        import gmpy2
    ...        return gmpy2.mpz('42')
    ...    def __mpq__(self):
    ...        import gmpy2
    ...        return gmpy2.mpq('13/12')
    >>> import gmpy2
    >>> gmpy2.mpq(A())
    mpq(13,12)
    >>> gmpy2.mpz(A())
    mpz(42)

casevh:
Thanks!

Slightly related question... When would you like me to release a 2.1 alpha release?

videlec:
This is on your side! But for me: as soon as possible!

Vincent K. has worked hard to experiment the integration in SageMath (see [ticket 22927](https://trac.sagemath.org/ticket/22927) and [ticket 22928](https://trac.sagemath.org/ticket/22928)). What we are missing:
 - direct constructors `GMPy_MPZ_From_mpz`, `GMPy_MPQ_From_mpq`, `GMPy_MPQ_From_mpz`
 - a gmpy2.pxd file with relevant declarations (to be installed at the same place as gmpy2.h)
 - robust header detection (ie what should the user do in her setup.py to find out where gmpy2.h/gmpy2.pxd are?)

casevh:
I would like to get an alpha release of 2.1 out soon. Most of the changes between 2.0 and 2.1 are related to support for thread-specific contexts instead of a global context. The mpz and mpq related code hasn't changed much. So while I expect bugs with mpfr and mpc, it should be reasonably stable for mpz and mpq.

I'll review your PR and commit it assuming it looks reasonable. (I'm travelling so my abilities to test are limited.)

PR #130, opened 2017-04-13
install header files

videlec:
Add a directive to install header files together with the library.

With these few lines added, the headers are installed in a directory `gmpy2/` inside the python installation.

Related to #121 

However, there would still remain the following problem: how can we find out where to find the gmpy2 repo that contains the header files?

PR #137, opened 2017-05-18
basic gmpy2.pxd and custom setup.py to install headers

videlec:
We have successfully tested the PR with our [pplpy](https://github.com/videlec/pplpy/) project. Sage integration is on its way.

You might want to add public interface to mpfr/mpc (see the TODO in `gmpy2.pxd` in commit b805459)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import os
 from distutils.core import setup, Extension
 from distutils.command.clean import clean
 from distutils.command.build_ext import build_ext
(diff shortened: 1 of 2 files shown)


casevh:
Thanks! I've started work on the public interface to mpfr/mpc. I'd like to get that into the first alpha release.

videlec:
Thanks. In my PR I made the instruction to install all of the .h files. You might want to include only gmpy2.h as mentioned in #130.

PR #138, opened 2017-05-19
Cython testsuite

videlec:
This PR fixes header installation (only installs gmpy2.h) and provides a testsuite for Cython.


casevh:
Thanks! I'm working on issue #139 next.

Issue #139, opened 2017-05-23
type checkers in the public C-API


videlec:
It would be convenient to have the following macros available from the public C-API
- `MPZ_Check` (now in `gmpy2_mpz.h`)
- `MPQ_Check` (now in `gmpy2_mpq.h`)
- `MPFR_Check` (now in `gmpy2_mpfr.h`)
- `MPC_Check` (now in `gmpy2_mpc.h`)

casevh:
I will try to add those tonight.

videlec:
Would be fantastic!

casevh:
After a review of the code, the macros are already available. See line ~500 in gmpy2.h.

Assume that we are just looking at MPZ_Check. The macro defined in gmpy2_mpz.h can directly reference the type object MPZ_Type. However, MPZ_Type is not directly accessible to an extension module and must be referenced via the GMPy_C_API array.

At least, that's what I think it does. Are you getting an error?

vinklein:
I just made a test, by adding MPZ_Check in gmpy2.pxd and calling it in a Cython pyx file, it works.

PR #140, opened 2017-05-24
add type checkers in gmpy.pxd

vinklein:
- add type checkers in gmpy.pxd
- add some tests in cython test suite
- Fix a bug, causing runtest.py to crash
if called from test_cython directory

related to issue #139


videlec:
You could also test that `MPZ_Check` return `False` on `mpq` (and vice versa). And also test with Python integers.

vinklein:
Ok. I've add tests for MPC_Check and MPFR_Check too.

PR #141, opened 2017-05-25
fix and tests Cython memory leak

videlec:
The following command in Cython

    py_obj = <object> my_py_obj

increases the reference counter of my_py_obj. As a consequence the following code gets wrong refcount

    res = <object> GMPy_MPZ_New(NULL)

 and the object `res` will never be deallocated by the garbage collector.

### DIFF
--- a/src/gmpy2.pxd
+++ b/src/gmpy2.pxd
@@ -1,5 +1,8 @@
 from cpython.object cimport PyObject
 
+cdef extern from "Python.h":
+    void Py_DECREF(PyObject *)
+
 cdef extern from "gmp.h":
     # gmp integers
     ctypedef struct __mpz_struct:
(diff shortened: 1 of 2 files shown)


casevh:
Thanks for keeping gmpy2.pxd up-to-date. Support for `mpfr` in the C-API should be done.

Interesting Cython behavior. How does Cython handle the following?

`res = <object> PyInt_FromLong(10)`

Or would you never do that...

videlec:
You can have a look at their headers [int.pxd](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/int.pxd). The C functions returning Python objects are declared as `object my_function(...)` which prevents this counter increment.

videlec:
> Support for mpfr in the C-API should be done.

I don't see `GMPy_MPFR_New` and `GMPy_MPC_New` in `src/gmpy2.h`. Is that intended?

casevh:
That was not intentional. I've committed a fix for `GMPy_MPFR_New`.

PR #145, opened 2017-06-06
Change --prefix to --shared in the README

sonderlain:
According to [this](https://github.com/aleaxit/gmpy/blob/c9e086e16d040d86a955900e4cd7593dadfab3f1/setup.py#L208) it's the "--shared" parameter what counts, not "--prefix". 
At least that was the reason I wasn't able to install the package.


casevh:
Thanks!

fengkx:
Please also change [here ](https://gmpy2.readthedocs.io/en/latest/intro.html#installing-gmpy2-on-unix-linux) at the blockquote.

PR #148, opened 2017-06-22
GMPy_MPFR_From_mpfr and more cython tests

videlec:
Better support for mpfr


casevh:
Thanks!

I will work on mpc support as soon as I can.

PR #159, opened 2017-10-05
Corrected bit_set description

TomsOzols:
The documentation for function/method "bit_set" is incorrectly portraying that the n-th bit would be changed to 0, while in fact it changes this bit to 1.


casevh:
Thanks!

Issue #156, opened 2017-08-24
mpq floor division


ces42:
floor division on `mpq`s just seems to return the numerator:

    Python 3.6.1 (default, Mar 22 2017, 06:17:05) 
    [GCC 6.3.0 20170321] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from gmpy2 import mpq
    >>> mpq(355, 113) // 2
    mpz(355)

is this intended?

casevh:
Not intended. I'll try to look at it this weekend.

videlec:
Could the issue be closed?

casevh:
Fix will be available in 2.1 series. The first release should be out soon.

ces42:
Thanks!

Issue #146, opened 2017-06-21
release?


videlec:
Is there any plan for alpha, beta, stable release?

Vincent K. already prepared all the integration in [pplpy](https://github.com/videlec/pplpy) and SageMath (see [ticket 22927](https://trac.sagemath.org/ticket/22927) and [ticket 22928](https://trac.sagemath.org/ticket/22928)). We are just waiting an official release to move on.

casevh:
I would still like to add mpc support to the C-API. I should be able to work on it this weekend. I should be able to make an alpha release by the end of June.

casevh:
I'm working on the mpc API and I think I spotted a bug in GMPy_MPFR_New(). Can you add a Cython test that uses a precision of 0? A precision of 0 should cause gmpy2 to use the precision of the currently active context. I suspect it may crash in the current gmpy2 code.

Note that contexts have subtly, but significantly, changed in version 2.1. In gmpy2 2.0.x, there was only a single global context that was shared by all thread. Contexts are now thread specific.

videlec:
Indeed, it does crash.

casevh:
I've added mpc to C API. GMPy_MPFR_New() and GMPy_MPC_New() shouldn't crash if you specify a precision of 0.

Can you also add a test where __mpz__ raises an exception? I think MPZ_Check() will crash if NULL is returned by PyObject_CallMethod().

vinklein:
I add mpc to gmpy.pxd, cython tests for mpc, and tests with 0 precision for mpfr and mpc and all is working fine.
Tests done where mpz raises exception doesn't cause crash.

isuruf:
@casevh, can we have official manylinux1 and osx wheels? I've tried building manylinux1 wheels and works fine. https://github.com/isuruf/gmpy2-wheels/releases. OSX wheels doesn't work yet.

videlec:
> I should be able to make an alpha release by the end of June.

Do you have a schedule?

casevh:
I believe all the critical issues for a first alpha have been resolved. I need to update the documentation and then I think I can make a source-only release soon. I will try for this weekend.

I will add a TODO file with the milestones for the following releases. At this point, I assume the next alpha release will follow soon.

casevh:
> can we have official manylinux1 and osx wheels? I've tried building manylinux1 wheels and works fine. https://github.com/isuruf/gmpy2-wheels/releases. OSX wheels doesn't work yet.

[...]

PR #160, opened 2017-10-09
Bug fix : mpq floor division #156

vinklein:
- Fix this bug. Appear when one of the parameters are not an MPQ (PyLong or MPZ for exemple).
- Add two test for these cases. 
- Fix needed for issue #146


casevh:
Thank you!

vinklein:
You're welcome !

PR #164, opened 2017-10-25
Add explicit cast to silence warnings on win64 platform

vinklein:
Add explicit cast to silence warnings on win64 platform.
For the four case in files gmpy2_convert_utils.c and gmpy2_convert_utils.c(195) where overflow are implossible. 
Issue #146


casevh:
Thank you!

PR #165, opened 2017-11-03
Fix win64 overflow for gmpy2_add(290) and gmpy2_add(347)

vinklein:
- related to issue #146 
- bug append on winx64 when you call add(mpfr(x), y) or add(y, mpfr(x)) with y > 2147483647
- add two doctest to test the fix


casevh:
I changed the logic to use long type instead of native_si since it was working with MPFR type. The native_si/native_ui should only be used with MPIR, not MPFR or MPC. My apologies for directing you down the wrong path.

vinklein:
No problem. Using the long type make more sense in these cases.

Issue #173, opened 2017-11-28
Segfault when objects have custom `__mpz__` method raising errors


videlec:
```
>>> import gmpy2
>>> class A:
...    def __mpz__(self): raise TypeError
>>> a = A()
>>> gmpy2.mpz(a)
** SEGFAULT **
```

vinklein:
Same problem with __mpq__, __mpfr__ and __mpc__ function.
All fixed in PR #174

casevh:
Thanks.

PR #174, opened 2017-11-28
gmpy2_cache.c : Fix issue #173

vinklein:
issue #173 
- Fix segfault when objects have custom `__mpz__`, `__mpq__`,
`__mpfr__` or `__mpc__` methods raising errors.

- add doctests for these cases.


casevh:
Thanks.

PR #180, opened 2018-01-25
Add data structures for the various types

jdemeyer:
Allow accessing the internals of the various types.

Second, by doing this, the function `GMPy_MPC_From_mpfr` can be used without linking against the MPC library (by replacing the `mpc_set_fr_fr` call by two `mpfr_set` calls). This makes sense because the function takes `mpfr_t` as input. So it's reasonable to expect linking against MPFR.


casevh:
Thanks!

Issue #179, opened 2018-01-24
mpc precision inconsistency


vinklein:
The precision of .real and .imag attributes looks to be inconsistent with their mpc's precision 
```
>>> from gmpy2 import *
>>> x = mpc('56+78j', precision=70)
>>> x.precision
(70, 70)
>>> x.real.precision
53
>>> x.imag.precision
53
```
Is this intended or is it a bug ?

casevh:
It's a bug. I'll work on a fix tomorrow.

vinklein:
Fixed with #181

PR #181, opened 2018-01-25
Fix for mpc precision inconsistency issue #179

vinklein:
Fix for issue #179 
- .real and .imag mpfr now return the same precision as their mpc
- add doctests for these cases.

### DIFF
--- a/src/gmpy2_mpc_misc.c
+++ b/src/gmpy2_mpc_misc.c
@@ -393,7 +393,10 @@ GMPy_MPC_GetImag_Attrib(MPC_Object *self, void *closure)
 
     CHECK_CONTEXT(context);
 
-    if ((result = GMPy_MPFR_New(0, context))) {
+    mpfr_prec_t rprec = 0, iprec = 0;
+    mpc_get_prec2(&rprec, &iprec, self->c);
+
+    if ((result = GMPy_MPFR_New(iprec, context))) {
         result->rc = mpc_imag(result->f, self->c, GET_MPFR_ROUND(context));
         _GMPy_MPFR_Cleanup(&result, context);
     }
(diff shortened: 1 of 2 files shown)


videlec:
Could you also test

- different precision for `real` and `imag`
- reconstruction
```
>>> x = gmpy2.mpc("1.3142123+4.3e-1001j", precision=(70,37))
>>> gmpy2.mpc(x.real, x.imag) == x
True
```

vinklein:
This one 
```
>>> gmpy2.mpc(x.real, x.imag) == x
True
```
don't work it return `False` because `gmpy2.mpc(..)` take the default precision if no precision is specified in parameter.

videlec:
> This one
>```
> >>> gmpy2.mpc(x.real, x.imag) == x
> True
>```
> don't work it return False because gmpy2.mpc(..) take the default precision if no precision is specified in > parameter.

That's a pity! In this very specific case the input has a very well defined precision that would better be used. But let's keep it for another PR.

vinklein:
On the other hand if you something like `gmpy2.get_context().real_prec = 100` you don't want it to be ignored if do an `gmpy.mpc()` call without `precision` paramater.

casevh:
Quick comment... try gmpy2,mpc(x.real, x.imag, precision=1) == x.

Specifying a precision of 1 should do an exact conversion if x.real and
y.real are either binary floating point (Python float or mpfr) or an
integer. At least some of the internal code paths do that. Since MPFR 4 now
considers 1 a valid precision, I want to change that magic value to -1.

It is possible to make the default value of precision be -1 instead of the
value in current context. This will allow lossless conversion by default.
You would need to specify a precision (or possibly pass a context) to get a
value converted to a different context. I personally like this option but
it is a backwards-incompatible change.

vinklein:
It doesn't work either.
```
>>> x = mpc("1.3142123+4.3e-1001j", precision=(70,37))
>>> mpc(x.real, x.imag, precision=1) == x
False
```
Is this a new issue ?

casevh:
It would be a new issue. I will merge this PR and create a new issue for maintaining precision of exact input.

PR #193, opened 2018-04-19
.travis.yml install a newer version of cython

vinklein:
With previous travis.yml cython was not detected by
test_cython/runtests.py and then no cython tests executed


casevh:
Thanks!

PR #196, opened 2018-06-06
Add gmpy2.h to manifest so it gets included in source distributions. …

### LINKED ISSUE #195: PyPI sources do not contain gmpy2/gmpy2.h symlink
The file `gmpy2/gmpy2.h` is not shipped in the source tarball. Because of this, the `gmpy2.h` header file is not installed when installing from the source tarball.

CC @pkittenis

Note: this was originally reported as #189 but without mentioning that this was sdist-specific.

pkittenis:
…Resolves #195

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,3 +5,4 @@ include docs/*
 include test/*
 include COPYING.LESSER
 include src/*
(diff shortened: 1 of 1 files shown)


casevh:
Thanks.

PR #201, opened 2018-06-27
Fix mpz += 0 setting the value to a random number.

jamesjer:
This fixes issue 200.


casevh:
Thanks for identifying the bug and finding the fix.That fragment of code was left over from some optimization testing and it was accidentally committed. I will do some performance tests and see if the optimization effort should be expanded or reverted.

Issue #204, opened 2018-10-01
2.1.0a3 missing files in source release


vinklein:
Hi case, 

Thanks for the source release  (#198).
Unfortunately pplpy doesnt compile with the package on pypi [gmpy2-2.1.0a3.tar.gz](https://files.pythonhosted.org/packages/46/72/3d7b689c618cdbc5353231d0cac82cb49c286f587c333e471607e4fb9328/gmpy2-2.1.0a3.tar.gz
), but it does with the current trunk.  Pypi's package seems to miss two files [gmpy2/gmpy2.pxd](https://github.com/aleaxit/gmpy/blob/master/gmpy2/gmpy2.pxd) and [gmpy/__init__.pxd](https://github.com/aleaxit/gmpy/blob/master/gmpy2/__init__.pxd).

pkittenis:
Those files seem to be missing from [MANIFEST.in](https://github.com/aleaxit/gmpy/blob/master/MANIFEST.in) which is the only way non-python files can be included into a source distribution.

vinklein:
@pkittenis Thanks for the tip.

vinklein:
@casevh 
Hi Case,
When do you plan to do another cython compatible source release  ?
May i help you in some way in order to speed up the process ?

casevh:
I have tagged the 2.1.0a4 release and the tar file that I plan to distribute can be downloaded from Github at https://github.com/aleaxit/gmpy/releases/tag/gmpy2-2.1a4

I will publish it on PyPi tomorrow.

vinklein:
Thanks !
I will do some tests today.

casevh:
I've published the latest source on PyPi.

PR #206, opened 2018-10-04
.travis.yml: Do ci tests on a tarball instead of ...

vinklein:
... the current trunk.
Instead of doing tests on the current trunk :
- Create a tarball.
- Install the tarball.
- Do python tests and cython tests with the installed tarball.

Testing the tarball can help to prevent issue like #204


casevh:
Thanks again! I'll release alpha 4 over the weekend.

Issue #213, opened 2019-03-04
Error during travis tests  "GMPY2 requires MPC 1.0.3 or later."


vinklein:
Current travis-ci tests failed on linux configuration (see: https://travis-ci.org/vinklein/gmpy/jobs/501590211)
```
 In file included from src/gmpy2.c:437:0:
  src/gmpy2.h:90:4: error: #error "GMPY2 requires MPC 1.0.3 or later."
   #  error "GMPY2 requires MPC 1.0.3 or later."
```

PR #216, opened 2019-03-05
travis.yml: Use xenial instead of trusty for  linux tests.

vinklein:
Trusty support will end next month.
Using xenial for test also fix issue #213


casevh:
Thanks!

Issue #242, opened 2019-05-22
qdiv modifies the first argument (regression in 2.1.0b1)


skirpichev:
I didn't debug things too much, but the problem could be easily reproduced in the [Diofant](https://github.com/diofant/diofant) repository.  See e.g. this build: https://travis-ci.org/diofant/diofant/builds/535696538.

Apparently, there is something wrong with the gmpy library, not with the application (i.e. diofant) itself: gmpy backend is optional and with builtin CPython's arithmetics everything is fine:
```
$ DIOFANT_GROUND_TYPES=python py.test -q diofant/tests/polys/test_rootisolation.py -k test_dup_sturm
.                                                                                                                               [100%]
====================================================== slowest 20 test durations ======================================================
0.13s setup    diofant/tests/polys/test_rootisolation.py::test_dup_sturm
0.01s call     diofant/tests/polys/test_rootisolation.py::test_dup_sturm

(0.00 durations hidden.  Use -vv to show these durations.)
1 passed, 9 deselected in 0.20 seconds
```
But
```
$ DIOFANT_GROUND_TYPES=gmpy py.test -q diofant/tests/polys/test_rootisolation.py -k test_dup_sturm
F                                                                                                                               [100%]
============================================================== FAILURES ===============================================================
___________________________________________________________ test_dup_sturm ____________________________________________________________

    def test_dup_sturm():
        R, x = ring("x", QQ)
    
        assert R.dup_sturm(5) == [1]
        assert R.dup_sturm(x) == [x, 1]
    
        f = x**3 - 2*x**2 + 3*x - 5
    
>       assert R.dup_sturm(f) == [f, 3*x**2 - 4*x + 3,
                                  -10*x/9 + QQ(13, 3), -QQ(3303, 100)]

diofant/tests/polys/test_rootisolation.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
diofant/polys/compatibility.py:398: in dup_sturm
    seq = dup_sturm(self.to_dense(f), self.domain)
diofant/polys/rootisolation.py:49: in dup_sturm
    s = dmp_rem(sturm[-2], sturm[-1], 0, K)
[...]

PR #245, opened 2019-05-26
Fix mutation in the first argument of qdiv()

skirpichev:
Two tests were altered.  Apparently, they were wrong:
before lines "gmpy2.qdiv(10, q)" and "q.__round__(4)" - q was
changed by previous calls to 2/5.

Closes aleaxit/gmpy#242

PS: Probably you could enable travis-ci for this repo and pull requests.  Anyway, this commit pass your CI tests, see [here](https://travis-ci.org/skirpichev/gmpy/builds/537415585), regression test was added.

### DIFF
--- a/src/gmpy2_mpq_misc.c
+++ b/src/gmpy2_mpq_misc.c
@@ -176,7 +176,7 @@ GMPy_MPQ_Function_Qdiv(PyObject *self, PyObject *args)
             goto arg_error;
         }
 
-        if (!(tempx = GMPy_MPQ_From_Rational(x, context)) ||
+        if (!(tempx = GMPy_MPQ_From_NumberAndCopy(x, context)) ||
             !(tempy = GMPy_MPQ_From_Rational(y, context))) {
             Py_XDECREF((PyObject*)tempx);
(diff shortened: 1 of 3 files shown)


casevh:
Thank you. I need to look at a couple of other issues and then I'll release beta 2.

skirpichev:
It seems, that gmpy2 now is noticeably slower, e.g. https://travis-ci.org/diofant/diofant/builds/537534305 (doesn't timeout on the Diofant's master).  Not sure if this is related to this bugfix.

casevh:
I think there may be a reference leak. I will look at it later today.

casevh:
There is no reference leak. There is no significant difference in performance for small arguments. It is possible that the tests are using very large values and the overhead for creating a copy become significant. I refactored the qdiv() code to not require copies but it require creating an extra mpq object. Can you test again with the current code?

skirpichev:
Sure, https://github.com/diofant/diofant/pull/840.  Build restarted: https://travis-ci.org/diofant/diofant/builds/537534305

skirpichev:
No luck.  I'll try to bisect regression on this week.

PR #254, opened 2019-10-16
Convert C file character encodings to UTF-8.

jamesjer:
Currently, src/gmpy2.c is ISO8859-1.  I don't recognize the encoding of mpz_pylong.c.  It might be doubly encoded.  In any case, this commit changes them both to Unicode.


casevh:
Thanks for the correction.

PR #258, opened 2020-01-31
Documentation for double_fac, multi_fac, primorial

sethtroisi:
I used
`grep --color -nriP '([^r]fac(?!(tor|ili|ts|tur))|primorial)' .`
to make sure that fac/doublefac/multifact/primorial appear in the same files


casevh:
Thanks!

PR #260, opened 2020-02-20
Add access to gmp limbs API

tylanphear:
This PR exposes the `mpz_limbs` [interface](https://gmplib.org/manual/Integer-Special-Functions.html) through 5 new methods on `xmpz` to provide access to the backing storage of an `xmpz` object. This is for fast access to the `gmp` internals to perform quick compression/decompression of the underlying `mpz` data.

Methods added:
- `xmpz.num_limbs() -> int`
  returns the number of `gmp` limbs underlying this `xmpz` object
- `xmpz.limbs_read() -> int`
  returns the address of an immutable buffer representing the limbs of `xmpz`
- `xmpz.limbs_write(n) -> int`
  potentially reallocates and returns the address of a mutable buffer representing the limbs of `xmpz`, sized to hold at least `n` limbs. May overwrite the existing limbs of `xmpz`.
- `xmpz.limbs_modify(n) -> int`
  potentially reallocates and returns the address of a mutable buffer representing the limbs of `xmpz`, sized to hold at least `n` limbs.
- `xmpz.limbs_finish(n) -> int`
  must be called after writing to the address returned by `xmpz.limbs(n)` to update the limbs of `xmpz` (and thereby see the changes take effect)

Attributes added:
- `xmpz.limb_size`
  Size of a `gmp` limb in bytes. Equal to `sizeof(mp_limb_t)`.

### DIFF
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -594,6 +594,7 @@ static PyObject *GMPyExc_Erange = NULL;
 #include "gmpy2_mpq_misc.c"
 #include "gmpy2_mpz_misc.c"
 #include "gmpy2_xmpz_misc.c"
+#include "gmpy2_xmpz_limbs.c"
 
 #ifdef VECTOR
 #include "gmpy2_vector.c"
@@ -1118,6 +1119,11 @@ PyMODINIT_FUNC initgmpy2(void)
 
     Py_INCREF(&XMPZ_Type);
     PyModule_AddObject(gmpy_module, "xmpz", (PyObject*)&XMPZ_Type);
+
+    PyObject* xmpz = XMPZ_Type.tp_dict;
+    PyObject* limb_size = PyIntOrLong_FromSize_t(sizeof(mp_limb_t));
+    PyDict_SetItemString(xmpz, "limb_size", limb_size);
+    Py_DECREF(limb_size);
 
     /* Add the MPQ type to the module namespace. */
 
--- a/src/gmpy2.h
+++ b/src/gmpy2.h
@@ -497,6 +497,7 @@ typedef struct {
 
 #include "gmpy2_xmpz_inplace.h"
 #include "gmpy2_xmpz_misc.h"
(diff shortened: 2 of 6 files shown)


casevh:
Thanks. This look good. I have a couple of comments.

Would you be okay with using the GMP names -- `xmpz.limbs_read()`, `xmpz.limbs_write` and so forth? I think that will make it easier to find the documentation. (And I had to read the source twice to notice the difference between `limbs()` and `limbs(n)`.)

Should it support both `limbs_write` and `limbs_modify` for completeness with matching the GMP API?

Do you have an example that could be modified for a test case?

tylanphear:
> Thanks. This look good. I have a couple of comments.
> 
> Would you be okay with using the GMP names -- `xmpz.limbs_read()`, `xmpz.limbs_write` and so forth? I think that will make it easier to find the documentation. (And I had to read the source twice to notice the difference between `limbs()` and `limbs(n)`.)
> 
> Should it support both `limbs_write` and `limbs_modify` for completeness with matching the GMP API?

I agree on both points. I've updated the branch as such.

> Do you have an example that could be modified for a test case?

My original use was in conjunction with `blosc`, specifically with `compress_ptr`/`decompress_ptr`. Obviously we wouldn't want a test dependency on `blosc`, so I'll have to do some thinking to come up with a different test case. Something like "create two xmpz objects, get the limbs of one, copy into the other, and then update and observe the change" should suffice.

tylanphear:
@casevh Ok, I've added a test that should cover the whole API.

nylonpsycho:
This is great! Just as a comment, I think it opens the door to looking at xmpz objects as binary faster than to_binary/from_binary, for non-portable use cases where speed is really important. I initially came across that here:

https://stackoverflow.com/questions/37101404/sharing-gmpy2-multi-precision-integer-between-processes-without-copying

casevh:
Looks great. Thanks!

PR #270, opened 2020-04-09
GitHub Action to pip install gmpy2 on multiple OSes and Python versions

cclauss:
Passes on macOS but fails on Linux and Windows  https://github.com/cclauss/gmpy/actions


casevh:
Thanks!

PR #273, opened 2020-04-18
Add .gitignore for .pyc files

martingkelly:
These get generated when you run the tests, but we shouldn't check them in,
so add them to .gitignore.


casevh:
Thanks!

casevh:
I just realized I never added my .gitignore file. It did exclude *.pyc and several other directories. I'll add the extra directories.

Issue #276, opened 2020-06-24
Big endian: wrong rounding mode


jamesjer:
We are having an issue with sympy on big endian machines in Fedora, which I have tracked down to this behavior.  On x86_64:
```
>>> gmpy2._mpmath_normalize(0, mpz(8658170283619779), -50, 53, 3, 'f')
(0, mpz(7), 0, 3)
```
On s390x:
```
>>> gmpy2._mpmath_normalize(0, mpz(8658170283619779), -50, 53, 3, 'f')
(0, mpz(1), 3, 1)
```
The issue is in the reading of the rounding mode.  This definition in gmpy2.h is the source of the problem:
```
#define Py2or3String_AsString       PyUnicode_AS_DATA
```
`PyUnicode_AS_DATA` was deprecated in python 3.3.  It takes a pointer to the internal character data and returns it cast to `char *`.  The problem is that the data is not really C `char` values; it is wide characters.  On little endian machines, when you dereference the return value of `PyUnicode_AS_DATA`, you get lucky and get the lowest significant byte of the first multibyte character, which happens to be the letter you want.  On big endian machines, you get unlucky and get the most significant byte of the first multibyte character, so the rounding mode is always read as the character `'\0'`.

PR #277, opened 2020-06-25
Correctly read rounding mode strings when characters are multibyte

jamesjer:
This is one approach that could be taken to resolve https://github.com/aleaxit/gmpy/issues/276.  It works for python 3.8 and 3.9, at least.  My reading of the docs suggests that it might work for versions as old as 3.3, but I haven't tested anything older than 3.8.


jamesjer:
I don't know what that windows failure is all about.  It didn't even get to the step where it attempted to compile gmpy, so I don't think it has anything to do with this change.

casevh:
Thanks for the bug report and the patch.

PR #283, opened 2020-11-13
Python 3.10 compatibility.

jamesjer:
The Fedora distribution has started doing experimental builds with alpha versions of python 3.10, to discover and fix problems in advance of the 3.10 release.  The gmpy code has 2 issues.  First, Py_SIZE() is no longer an lvalue in 3.10.  A Py_SET_SIZE macro was introduced in 3.9 in anticipation of this change; its use is mandatory in 3.10.  The second issue is test code that checks whether the python version string starts with "3.1"; that is true of 3.10 also, of course.  Since setup.py claims support only for 3.4 and later, it may be better to simply remove that entire block of code.  Let me know if you prefer that and I will update this commit.

### DIFF
--- a/src/gmpy2_convert_gmp.c
+++ b/src/gmpy2_convert_gmp.c
@@ -222,10 +222,18 @@ GMPy_PyLong_From_MPZ(MPZ_Object *obj, CTXT_Object *context)
     while ((size>0) && (result->ob_digit[size-1] == 0)) {
         size--;
     }
+#if PY_VERSION_HEX >= 0x030900A4
+    Py_SET_SIZE(result, size);
+#else
     Py_SIZE(result) = size;
+#endif
 
     if (negative) {
+#if PY_VERSION_HEX >= 0x030900A4
+        Py_SET_SIZE(result, - Py_SIZE(result));
+#else
         Py_SIZE(result) = - Py_SIZE(result);
+#endif
     }
     return (PyObject*)result;
 }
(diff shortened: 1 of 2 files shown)


src/gmpy2_convert_gmp.c
```diff
@@ -222,10 +222,18 @@ GMPy_PyLong_From_MPZ(MPZ_Object *obj, CTXT_Object *context)
     while ((size>0) && (result->ob_digit[size-1] == 0)) {
         size--;
     }
+#if PY_VERSION_HEX >= 0x030900A0
```
hroncok, line 225:
```suggestion
#if PY_VERSION_HEX >= 0x030900A4
```

This from https://docs.python.org/3.10/whatsnew/3.10.html

src/gmpy2_convert_gmp.c
```diff
@@ -222,10 +222,18 @@ GMPy_PyLong_From_MPZ(MPZ_Object *obj, CTXT_Object *context)
...
+    Py_SET_SIZE(result, size);
+#else
     Py_SIZE(result) = size;
+#endif
 
     if (negative) {
+#if PY_VERSION_HEX >= 0x030900A0
```
hroncok, line 232:
```suggestion
#if PY_VERSION_HEX >= 0x030900A4
```

hroncok (commented):
A small nitpick. Not many people would use 3.9 alphas now, but this was introduced in 3.9.0a4.

src/gmpy2_convert_gmp.c
jamesjer: Good catch.  I've updated the commit as suggested.

casevh:
Thanks. I was just getting ready to test with 3.10.

PR #284, opened 2020-11-21
Added prev_prime(x)

sethtroisi:
I authored `mpz_prevprime` and finally got it upstreamed. I'd love if gmpy2 would also support this (as I mostly call gmp via your wonderful wrapper).

Couple of notes
* I throw `Py_RETURN_NOTIMPLEMENTED` if `__GNU_MP_VERSION < 6.3`
I'm not sure when gmp 6.3 release will happen but this is the safest option. I/you/others can always remove this line (or change it to 6.2) when using development version of the repo which uses build version 6.2.99.
* mpz_prevprime has an int return (0 when no previous prime (arg < 2), definite prime, probable prime) I can return a two element tuple but IMHO it's better to just return the previous_prime (or 2).

### DIFF
--- a/docs/mpz.rst
+++ b/docs/mpz.rst
@@ -105,7 +105,19 @@ mpz Functions
 .. autofunction:: powmod_exp_list
 .. autofunction:: powmod_base_list
 .. autofunction:: powmod_sec
-.. autofunction:: prev_prime
+.. only:: has_prev_prime
+
+   ..
+       Can't use autofunction:: prev_prime as sphinx tries to import
+       regardless of has_prev_prime conditional.
+
+
+   .. function:: prev_prime(x, /) -> mpz
+
+      Return the previous *probable* prime number < x.
+
+      Only present when compiled with GMP 6.3.0 or later.
+
 .. autofunction:: primorial
 .. autofunction:: remove
 .. autofunction:: t_div
--- a/docs/mpz.rst
+++ b/docs/mpz.rst
@@ -105,6 +105,7 @@ mpz Functions
 .. autofunction:: powmod_exp_list
 .. autofunction:: powmod_base_list
 .. autofunction:: powmod_sec
+.. autofunction:: prev_prime
 .. autofunction:: primorial
 .. autofunction:: remove
 .. autofunction:: t_div
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -781,7 +781,9 @@ static PyMethodDef Pygmpy_methods [] =
     { "mul", GMPy_Context_Mul, METH_VARARGS, GMPy_doc_function_mul },
     { "multi_fac", GMPy_MPZ_Function_MultiFac, METH_VARARGS, GMPy_doc_mpz_function_multi_fac },
     { "next_prime", GMPy_MPZ_Function_NextPrime, METH_O, GMPy_doc_mpz_function_next_prime },
+#if (__GNU_MP_VERSION > 6) || (__GNU_MP_VERSION == 6 &&  __GNU_MP_VERSION_MINOR >= 3)
     { "prev_prime", GMPy_MPZ_Function_PrevPrime, METH_O, GMPy_doc_mpz_function_prev_prime },
+#endif
     { "numer", GMPy_MPQ_Function_Numer, METH_O, GMPy_doc_mpq_function_numer },
     { "num_digits", GMPy_MPZ_Function_NumDigits, METH_VARARGS, GMPy_doc_mpz_function_num_digits },
     { "pack", GMPy_MPZ_pack, METH_VARARGS, doc_pack },
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -406,7 +406,7 @@
  *    Adjust test suite to reflect changes in output in MPFR 4.1.0.
  *
  *    2.1.0b6
- *    Improve argument type processing by saving type information to 
+ *    Improve argument type processing by saving type information to
  *        decrease the number of type check calls. Especially helpful
  *        for mpfr and mpc types. (Not complete but common operations
  *        are done.)
@@ -474,7 +474,8 @@
  *    gcd()/lcm() now uses vectorcall protocol. (skirpichev)
  *    Expose context type. (skirpichev)
  *    Correct error in is_strong_bpsw_prp. (casevh)
- *    
+ *    Added prev_prime if part of loaded gmp
+ *
  ************************************************************************
  *
  * Discussion on sizes of C integer types.
@@ -780,6 +781,7 @@ static PyMethodDef Pygmpy_methods [] =
     { "mul", GMPy_Context_Mul, METH_VARARGS, GMPy_doc_function_mul },
     { "multi_fac", GMPy_MPZ_Function_MultiFac, METH_VARARGS, GMPy_doc_mpz_function_multi_fac },
     { "next_prime", GMPy_MPZ_Function_NextPrime, METH_O, GMPy_doc_mpz_function_next_prime },
+    { "prev_prime", GMPy_MPZ_Function_PrevPrime, METH_O, GMPy_doc_mpz_function_prev_prime },
     { "numer", GMPy_MPQ_Function_Numer, METH_O, GMPy_doc_mpq_function_numer },
     { "num_digits", GMPy_MPZ_Function_NumDigits, METH_VARARGS, GMPy_doc_mpz_function_num_digits },
     { "pack", GMPy_MPZ_pack, METH_VARARGS, doc_pack },
--- a/src/gmpy2_mpz_misc.c
+++ b/src/gmpy2_mpz_misc.c
@@ -1505,12 +1505,13 @@ GMPy_MPZ_Function_NextPrime(PyObject *self, PyObject *other)
 
 PyDoc_STRVAR(GMPy_doc_mpz_function_prev_prime,
 "prev_prime(x, /) -> mpz\n\n"
-"Return the previous *probable* prime number < x.");
+"Return the previous *probable* prime number < x.\n"
+"Only present when compiled with GMP 6.3.0 or later.");
 
+#if (__GNU_MP_VERSION > 6) || (__GNU_MP_VERSION == 6 &&  __GNU_MP_VERSION_MINOR >= 3)
 static PyObject *
 GMPy_MPZ_Function_PrevPrime(PyObject *self, PyObject *other)
 {
-    #if (__GNU_MP_VERSION > 6) || (__GNU_MP_VERSION == 6 &&  __GNU_MP_VERSION_MINOR >= 3)
         MPZ_Object *result;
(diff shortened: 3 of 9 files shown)


casevh:
Hi Seth,

Congratulations for getting prev_prime() accepted. I will merge this in a few days.

sethtroisi:
I inverted the `#if` which fixes compilation. Tests are still broken because `libgmp-dev` is version 6.2
Maybe I should comment out the tests and leave a note to enable them in 6months/1year when 6.3 is released?

casevh:
I support two different versions of MPFR and then split the tests into
separate files. I'll review this once I get 2.1.0 released.

sethtroisi:
libgmp now contains prevprime in 6.2.1! I'll take a look at fixing the conflicts and getting this back in shape for merging.

casevh:
I've generally used conditional compilation to only include a new function if the underlying library provides it. But I may have another option for including new features. Don't update your PR just yet.

skirpichev:
@casevh, I think it's time to review this (gmp-6.3.0 is available).

@sethtroisi, could you, please, fix merge conflicts?  Documentation structure was changed, let me know if you will need some help.

sethtroisi:
I think CI requires than #425 is merged. Currently GMP 6.3 isn't downloaded which results in `PrevPrime` returning `NotImplemented` in the tests.

src/gmpy2_mpz_misc.c
```diff
@@ -1503,6 +1503,46 @@ GMPy_MPZ_Function_NextPrime(PyObject *self, PyObject *other)
...
 }
 
+PyDoc_STRVAR(GMPy_doc_mpz_function_prev_prime,
+"prev_prime(x, /) -> mpz\n\n"
+"Return the previous *probable* prime number < x.");
+
+static PyObject *
+GMPy_MPZ_Function_PrevPrime(PyObject *self, PyObject *other)
+{
+    #if (__GNU_MP_VERSION > 6) || (__GNU_MP_VERSION == 6 &&  __GNU_MP_VERSION_MINOR >= 3)
```
skirpichev, line 1513: Given f8878b6 (and 39a9e48), I would guess, that you should surround the whole function by ``#if``'s.

test/test_gmpy2_mpz_misc.txt
```diff
@@ -652,9 +652,32 @@ TypeError:
...
+mpz(2)
+>>> gmpy2.prev_prime(1000004)
+mpz(1000003)
+>>> gmpy2.prev_prime(1000033)
+mpz(1000003)
+>>> gmpy2.prev_prime(-100)
+mpz(2)
+>>> gmpy2.prev_prime(1)
+mpz(2)
+
```
skirpichev, line 680: Per #377, you shouldn't use test/*.txt files for new tests.

These tests could be moved to the test_mpz.py, I think.  And you could use the pytest's skipif fixture to skip them if the GMP library is too old.

docs/mpz.rst
```diff
@@ -105,6 +105,7 @@ mpz Functions
 .. autofunction:: powmod_exp_list
 .. autofunction:: powmod_base_list
 .. autofunction:: powmod_sec
+.. autofunction:: prev_prime
```
skirpichev, line 108: Not sure if we should include this function to the sphinx docs.

But if so, you should do this conditionally, probably with the ["only" directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-only) of the sphinx.  Also, I think that in this case we should mention in the prev_prime docstring, that this function is available only for certain conditions.

src/gmpy2_mpz_misc.c
```diff
@@ -1503,6 +1503,46 @@ GMPy_MPZ_Function_NextPrime(PyObject *self, PyObject *other)
...
+                mpz_set_ui(result->z, 2);
+            }
+        }
+        else {
+            if (!(result = GMPy_MPZ_From_Integer(other, NULL))) {
+                TYPE_ERROR("prev_prime() requires 'mpz' argument");
+                return NULL;
+            }
+            else {
+                if (!mpz_prevprime(result->z, result->z)) {
```
skirpichev, line 1533: Docs says in this case the rop is unchanged.  Maybe you should return result without any tests?

skirpichev:
> I think I'm testing the correct thing by using
```
pip install .
python test/runtests.py
```
Should be enough.

> I have some test failures but they are certainly related to mpc e.g. mpc('nan+nanj') does not match mpc('nannanj').

Probably related to #422. If you did a rebase wrt the current master - there should be no failures.

> I think CI requires than https://github.com/aleaxit/gmpy/pull/425 is merged. Currently GMP 6.3 isn't downloaded which results in PrevPrime returning NotImplemented in the tests.

Yes, #425 is required.  But tests should pass in any case.

src/gmpy2_mpz_misc.c
casevh: I prefer to only include functions that are supported by the GMP/MPFR/MPC version. I think it causes less confusion than raising NonImpemented. The documentation can include prev_prime with a note "Only present when compiled with GMP 6.3.0 or later."

sethtroisi: I believe the documentation is in the build artifacts and I'll check that it contains that note which I added to `PyDoc_STRVAR(GMPy_doc_mpz_function_prev_prime`.

test/test_gmpy2_mpz_misc.txt
sethtroisi: Done

src/gmpy2_mpz_misc.c
sethtroisi: I changed it to ValueError.

src/gmpy2_mpz_misc.c
```diff
@@ -1503,6 +1503,46 @@ GMPy_MPZ_Function_NextPrime(PyObject *self, PyObject *other)
...
+    #if (__GNU_MP_VERSION > 6) || (__GNU_MP_VERSION == 6 &&  __GNU_MP_VERSION_MINOR >= 3)
+        MPZ_Object *result;
+
+        if(MPZ_Check(other)) {
+            if(!(result = GMPy_MPZ_New(NULL))) {
+                /* LCOV_EXCL_START */
+                return NULL;
+                /* LCOV_EXCL_STOP */
+            }
+            if (!mpz_prevprime(result->z, MPZ(other))) {
```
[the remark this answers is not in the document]
sethtroisi: This matches the style of the `next_prime` code directly above it. I tried changing to `other->z` but that fails here.

test/test_gmpy2_mpz_misc.txt
skirpichev: Some *.txt are there.

src/gmpy2.c
```diff
@@ -780,6 +781,7 @@ static PyMethodDef Pygmpy_methods [] =
     { "mul", GMPy_Context_Mul, METH_VARARGS, GMPy_doc_function_mul },
     { "multi_fac", GMPy_MPZ_Function_MultiFac, METH_VARARGS, GMPy_doc_mpz_function_multi_fac },
     { "next_prime", GMPy_MPZ_Function_NextPrime, METH_O, GMPy_doc_mpz_function_next_prime },
+    { "prev_prime", GMPy_MPZ_Function_PrevPrime, METH_O, GMPy_doc_mpz_function_prev_prime },
```
skirpichev, line 784: This you should do also conditionally.  Ditto in [src/gmpy2_mpz_misc.h](https://github.com/aleaxit/gmpy/pull/284/files#diff-98ae8bc1d201bda720b449caed6fa3091db148f52c914b5a959604b9e91fabf0)

src/gmpy2_mpz_misc.c
skirpichev: > I believe the documentation is in the build artifacts

No.  The documentation build will fail, unless you exclude new function for GMP < 6.3.0.

src/gmpy2_mpz_misc.c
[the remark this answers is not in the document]
skirpichev: Ok.

> I tried changing to other->z but that fails here.

`other` is a ``PyObject*``, the MPZ macro do typecast.

src/gmpy2_mpz_misc.c
skirpichev: OR (perhaps, it's better), you can include the ``prev_prive`` (a copy of the docstring) to the sphinx docs with the ``function`` directive.

docs/mpz.rst
skirpichev:
```suggestion
.. function:: prev_prime(x, /) -> mpz

   Return the previous *probable* prime number < x.
   Only present when compiled with GMP 6.3.0 or later.

```

src/gmpy2_mpz_misc.c
sethtroisi: I got sphinx running locally and verified that the prev_prime is included (conditional on has_prev_prime which is hardcoded in conf.py as `mp_version() >= "GMP 6.3.0"`)

test/test_gmpy2_mpz_misc.txt
sethtroisi: I left one additional test for next_prime(1000000) which I think is a nice test I can remove if you don't want it added here.

docs/mpz.rst
sethtroisi: It took me a few days but I made a conditional tag and got this working with the only directive (thanks for pointing me at it).

skirpichev: The problem with your solution is that this function will not be displayed on the https://gmpy2.readthedocs.io/en/latest/ for a long time.  That's why I did this suggestion.

sethtroisi: Yep you were persepctive in a way I had to learn by making the mistake. Should be fixed now. I tried to run build, tests, and sphinx, with and without gmp 6.3.0.

test/test_mpz_functions.txt
```diff
@@ -523,6 +523,9 @@ Test gmpy2.mul_2exp
 Test gmpy2.next_prime
 ---------------------
 
+Test gmpy2.prev_prime
+---------------------
+
```
skirpichev, line 528:
```suggestion
```

sethtroisi:
Here I was writing a long question about if anything more was needed. :) Thanks for pulling I'm glad to have gotten this included!

skirpichev:
> Here I was writing a long question about if anything more was needed. :) 

@sethtroisi, as I said [above](https://github.com/aleaxit/gmpy/pull/284#discussion_r1319220394) - you could just use "function" directive without any checks.  Else the sphinx docs on the https://gmpy2.readthedocs.io/en/latest/ or in the build artifact will not include the ``prev_prime()`` function.  @casevh, probably we *want* to include this unconditionally (with a warning), right?

skirpichev:
See #430

Issue #291, opened 2021-02-21
Allow multithreading during mpz operations?


jepler:
Isn't it safe to allow threads during long-running mpz operations?  For instance
```c
diff --git a/src/gmpy2_mul.c b/src/gmpy2_mul.c
index e312769..e147508 100644
--- a/src/gmpy2_mul.c
+++ b/src/gmpy2_mul.c
@@ -45,7 +45,9 @@ GMPy_Integer_MulWithType(PyObject *x, int xtype, PyObject *y, int ytype,
 
     if (IS_TYPE_MPZANY(xtype)) {
         if (IS_TYPE_MPZANY(ytype)) {
+            Py_BEGIN_ALLOW_THREADS;
             mpz_mul(result->z, MPZ(x), MPZ(y));
+            Py_END_ALLOW_THREADS;
             return (PyObject*)result;
         }
```

This is of interest in a program of mine, as individual operations are slow but I have a number of them to do in parallel.  Making this change locally (and to add) lets me get about a 7x speed-up on my particular workload.

Am I overlooking something about how this would interact with the Python GIL? If not, is there any interest in a PR along these lines?

casevh:
I looked at this a few years ago. IIRC, the test results were unpredictable with threading but predictable with multiple process. My test was easily done on multiple processes so it really isn't fair. And my changes to the threading model have been make. I think it is worth trying again. 

I'd like to avoid slowing down the fast operations. A couple of ideas:

1. Use a context flag to enable threading support (release_gil ??) with a default of False. 
2. To avoid releasing the GIL for small operands, we may want to add a check for the length of the underlying mpz_t.

I like the option of a context flag that defaults to False for the soon-to-be-released 2.1.0 version.

A PR would be great.

casevh:
Hi,

I will add this to the next release. I need to make a final release of the current code and then consider it a bug-fix only version. I don't want to add a change of this scope this late in the process. 

Case

jepler:
That's excellent, thank you! Immediately after filing this issue I second-guessed whether it could be as simple as it seemed.

I have found that for my particular "application" (computing large numbers for their own sake, nothing practical) that there is a range of numbers where, particularly for multiplication, it helps to perform multiplication in a thread with ThreadPool.imap. Below the threshold the individual operations are too fast, and above the threshold the numbers are big enough that they don't fit in RAM and swapping begins. With the process model, transferring the representation of the number appears to have an impact on performance.

[...]

PR #293, opened 2021-03-03
Allow threads in certain mpz operations, subject to a flag

jepler:
As discussed in #291 this allows threads during certain MPZ operations.  The choice is made based on a new flag in the context structure, but not on the characteristics (size) of the numbers involved.

Testing performed:
 * built on Linux with Python 3.7.3
 * Ran `tests/runtests` (some expected output updated)
 * Ran my [fib programs](https://gist.github.com/f3566b4d9222fd29839f1260bbcc0b01), comparing threaded and unthreaded performance in the range where threads are a benefit:
```
$ time python3 fib6.py $((1<<27))  # Threads enabled
134217728: 28049846 digits: 803430834915283969401079…337223560039659570859461

real	0m1.749s
user	0m6.196s
sys	0m0.411s
$ time python3 fib5.py $((1<<27)) # Threads disabled
134217728: 28049846 digits: 803430834915283969401079…337223560039659570859461

real	0m5.777s
user	0m5.386s
sys	0m0.312s
```

The only threaded operation my program performs is "*" of mpz long integers.

I'll look into any CI failures that may occur, and I can add the new begin/end thread macros in other sites you think would be useful to general users. I'm also not sure if there are other testing or documentation tasks I should perform.

I had trouble rebuilding things when I changed the C source.  As I'm not particularly familiar with Python packaging, I expected 'setup.py build' to re-build things, but it does not seem to noticed changed source files. Can you let me know what I'm doing wrong?

### DIFF
--- a/src/gmpy2.h
+++ b/src/gmpy2.h
@@ -171,6 +171,7 @@ typedef struct {
     mpfr_rnd_t imag_round;   /* current rounding mode for Im(MPC) */
     int allow_complex;       /* if 1, allow mpfr functions to return an mpc */
     int rational_division;   /* if 1, mpz/mpz returns an mpq result */
+    int allow_release_gil;   /* if 1, allow mpz functions to release the GIL */
 } gmpy_context;
 
 typedef struct {
--- a/src/gmpy2_add.c
+++ b/src/gmpy2_add.c
@@ -46,7 +46,9 @@ GMPy_Integer_AddWithType(PyObject *x, int xtype, PyObject *y, int ytype,
 
     if (IS_TYPE_MPZANY(xtype)) {
         if (IS_TYPE_MPZANY(ytype)) {
+            GMPY_MAYBE_BEGIN_ALLOW_THREADS(context);
             mpz_add(result->z, MPZ(x), MPZ(y));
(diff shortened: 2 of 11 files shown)


casevh:
Thanks! I want to try a couple experiments. Don't be surprised if I temporarily revert the change until I get 2.1.0 released. 

The magic build command is:

python setup.py build_ext --force install --force

jepler:
Thank you!  Whether it ends up reverted or not, I appreciate your attentiveness to my issue & PR.

casevh:
Just to clear, if I revert the PR, it will be recommitted as soon as 2.1.0 becomes official. I plan to support releasing the GIL. I'll have more comments tonight.

casevh:
I've made a few minor changes. The "CHECK_CONTEXT(context)" is now done at the top of the function and has been removed from the macro. This aligns with the MFPR and MPC related functions. So the GIL is now released for more argument combinations including mpz, mpq, mpfr, and mpc. I haven't written any threading specific tests or demos yet.

jepler:
Cool. I don't know if it's the kind of demo you're interested in but I'm happy to provide an appropriately-licensed version of my fib program, though I don't think it's really reflective of typical usage.

PR #297, opened 2021-05-14
Use _Py_HashPointer for python 3.10

vaartis:
In python 3.10 _PyHASH_NAN was removed and its usage replaced with _Py_HashPointer.

see https://github.com/python/cpython/blob/3.10/Python/pyhash.c#L102


skirpichev:
FYI, bug in the RH: https://bugzilla.redhat.com/show_bug.cgi?id=1959010

casevh:
Thank you.

Issue #192, opened 2018-04-19
two obsolete mpfr functions


videlec:
In the last release of mpfr there are two deprecated functions that appear to be used by gmpy2
```
src/gmpy2_random.c: In function 'GMPy_MPFR_grandom_Function':
src/gmpy2_random.c:248:5: warning: 'mpfr_grandom' is deprecated

src/gmpy2_math.c: In function 'GMPy_Real_Root':
src/gmpy2_math.c:1148:1: warning: 'mpfr_root' is deprecated
```

casevh:
I've looked at the MPFR documentation and I can rewrite the gmpy2 functions using the replacement mpfr functions. There will be a slight difference in behavior of root(): the sign of root(-0, k), k is even, becomes positive instead of negative. The new behavior matches the IEEE7 54-2008 standard.

Should I enforce the IEEE 754-2008 behavior even when using MPFR 3?

PR #300, opened 2021-05-27
Remove calls of obsolete mpfr functions

skirpichev:


skirpichev:
Failure: Ubuntu 16.04 doesn't have MPFR >= 4.  Let me know if it does make sense to support this release.

skirpichev:
Ok, I've added "#if" block.

skirpichev:
@casevh, does this make sense for you?

casevh:
The functions you've changed are already guarded by a version check and they are only used with MPFR v3.x. There is a lot of duplicated code but that's a different issue.

Is there a scenario where compilation fails on an old Linux distribution?

I'll test on Ubuntu 16.04 again but it was fine the last time I checked.

skirpichev:
> The functions you've changed are already guarded by a version check and they are only used with MPFR v3.x.

@casevh, yes.  But the #192 was about removing warnings from these checks.

The only problem I see is a different sign conventions for mpfr_root() and mpfr_rootn_ui().

> Is there a scenario where compilation fails on an old Linux distribution?

I doubt.  For old MPFR we fallback on the old code.

skirpichev:
Ok, I think this should be closed.

Issue #210, opened 2019-02-21
gmpy2 can't parse valid Python number literals


skirpichev:
I expect, that any [correct number](https://docs.python.org/3/reference/lexical_analysis.html#numeric-literals), coming in `str` form must be accepted by gmpy2's mpz/mpq/mpfr.

Unfortunately, this is not the case.  For example, underscores can't be used to group digits (spaces sometimes are accepted instead):
```
Python 3.7.2+ (default, Feb  2 2019, 14:31:48) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gmpy2
>>> gmpy2.version()
'2.1.0a4'
>>> gmpy2.mpz('1 2 3')
mpz(123)
>>> int('1 2 3')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1 2 3'
>>> int('1_2_3')
123
>>> gmpy2.mpz('1_2_3')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid digits
>>> float('1.2_3_4')
1.234
>>> gmpy2.mpfr('1.2_3_4')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid digits
>>> gmpy2.mpfr('1.2 3 4')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid digits
```

casevh:
I'm testing a fix for mpz(). I'll fix mpfr() next. The use of spaces as separators is a GMP behavior, no MPFR. Underscores were introduces in Python 3.6.

casevh:
I've done some testing and it becomes complicated to exactly match match Python's handling of underscores in all cases. I tried the simple approach of just removing the underscore characters but then strings that are invalid for Python become valid in gmpy2. (int('1_2_') raises an exception in Python but returns mpz(12) with gmpy2.)

If the expectation is that valid Python strings are converted to the same value, then just deleting underscores should work, If all the expectation is that all invalid Python strings raise an exception, then I don't think that will ever be possible.

skirpichev:
On Sun, Mar 10, 2019 at 09:25:32AM -0700, casevh wrote:
>    If the expectation is that valid Python strings are converted to the same
>    value, then just deleting underscores should work, If all the expectation
>    is that all invalid Python strings raise an exception, then I don't think
[...]

PR #301, opened 2021-05-28
Support parsing of CPython-style numeric literals

skirpichev:
I'm less certain about the second commit (56afdc1).  There is no a direct analog of mpfr() in the CPython stdlib.  Thus, perhaps we shouldn't support here CPython-style parsing...


casevh:
The fallback to int() when mpz() fails is okay, except slower, since it does not lose precision. I don't want the loss of precision by converting to a float first.

I looked at the code in _Py_string_to_number_with_underscores. I think the logic can be converted to support mpz/mpq/mpfr/mpc conversions. For mpz(), I would change the logic for "underscores only allowed after decimal digits" to either rely on the specified base or just allow digits and all upper/lower case ASCII characters. I'll give it a try tonight or tomorrow.

skirpichev:
> I don't want the loss of precision by converting to a float first.

Well, maybe we should use less strict approach.  In fact, the Decimal class of stdlib - does violate the PEP 515 (https://bugs.python.org/issue44267) and this is ok...  So, probably we can just strip underscores in case of mpfr.

As I told you before, maybe all this stuff belongs to gmp/mpfr.  Other projects implement similar workarounds for string format restrictions of mpz/mpq/etc.  For example, https://github.com/ocaml/Zarith/pull/75.  What do you think?

skirpichev:
Ok, the second commit was replaced with a more simple solution, mentioned before.

casevh:
I'm working on an updated change based on the approach you used for mpfr. I'm just removing all underscore characters. It it working but ...

Unrelated, I'm tracking down some reference leaks.

skirpichev:
> I'm just removing all underscore characters.

That's a bad approach in general (PEP violation).  For mpfr() it might be fine due to mentioned above reason.  But not for int's.

skirpichev:
Ok, @casevh, apparently this pr is not relevant anymore.  I still think, however, that for mpz() your solution with replace() is wrong (violation of the PEP 515).  But feel free to close related issue #210, if you disagree.

PR #347, opened 2022-11-04
Use pyproject.toml for metadata and drop support for old CPython versions

skirpichev:


casevh:
I will commit this after I release v3.1.3. I'll update the version to 3.2 at that time also.

skirpichev:
@casevh, maybe you can do a minor release with wheels for the 3.11?

casevh:
I think I did something wrong with the pull request. There was a conflict with setup.py that I tried to fix. I'm not certain what I did wrong. Git still confuses me at times. All the test builds broke.

I will just merge it and assume I just need to fix setup.py.

If I need help, I'll add a comment.

casevh:
Fixed setup.py and local build is working. Updated version to 2.2.0a1 and pushed.

Thank you for this change.

Case

PR #355, opened 2023-01-23
Minor fixes for docs

skirpichev:
* README -> README.rst
* fix description for some instance attributes
* remove redundant/useless sphinx directives
* fix sphinx warning for docs/win_build.rst
* enable docs building in GA
* Use sphinx_rtd_theme for local builds too

PR #356, opened 2023-01-24
Use sphinx's autodoc & intersphinx extensions

skirpichev:
Rationale: It seems for me, that the rst documentation for the gmpy2 functions/classes just reproduce docstrings without extra rst-formatting and so on.  Hence, we can just reuse plain docstrings from the code with the autodoc sphinx extension (which is enabled, but not used so far).

So far, only classes/functions in the mpz.rst were converted, just to show how it looks (the pip_install_gmpy2 has the build artifact with html and pdf docs after #355).  If you feel this does make sense - please decide how this should be merged: I can convert all docs in one shot (this pr) or gradually with several prs.

Note: In this example I use ability to use the stdlib types as generic types (e.g. to show a return value of 2-type `tuple[mpz, mpz]`).  That's coming in the CPython 3.9 (it's present in the Debian stable, bullseye - as an example of a conservative linux distribution) with the [PEP 585](https://peps.python.org/pep-0585/).  In case you are planning to support more older Python 3.x versions - that part should be reverted.


casevh:
I'm assuming the requirement for Python 3.9+ only impacts building the documentation. I'm fine with requiring Python 3.9+ for building the official documentation as long as previous versions either fail with a friendly error message or create documentation with a different format.

Thanks for the work on the documentation cleanup. It's been on my list but I don't have much free time lately.

Case

skirpichev:
Ok, then I'll continue with the rest of docs.

sethtroisi:
When I run `sphinx-build --color -W --keep-going -b doctest docs build/sphinx/doctest` on a clean checkout under python3.8 with Sphinx 7.1.2 I get some opaque errors

`docstring of builtins.xmpz:16: WARNING: Inline emphasis start-string without end-string.`
and `WARNING: autodoc: failed to import function 'gamma_inc' from module 'gmpy2'; the following exception was raised:`

When I switched over to python3.10 and Sphinx 7.2.5 these errors go away. Adding a friendly error message would be nice for us with system defaults of python3.8

skirpichev:
@sethtroisi, as I said above: minimal requirements for building docs is the python3.9.  We definitely could check that in docs/conf.py and raise a RuntimeError.

sethtroisi:
Thanks for adding the check in https://github.com/skirpichev/gmpy/commit/dabcb8d29280c0e01058a950cb981922a9716a87

PR #359, opened 2023-01-26
Remove python2-specific code

skirpichev:
* from \_\_future\_\_ imports
* py2-related tests
* py2-related VS build files, etc

I would suggest to remove also build.vc* and archive directories.  Git will keep the history anyway...

IIUC, windows_build.txt and docs/win_build.rst are obsoleted too and the preferred way to build the project on the windows is the mingw64/msys2_build.txt.  If this is correct - we could point users to the mingw64/msys2_build.txt and remove other instructions.


casevh:
This looks good to me. Thanks for the work you're doing.

Case

PR #360, opened 2023-01-26
Remove archive/ and build.vc*/ directories

skirpichev:
a continuation of #359


casevh:
Thanks for the cleanup.

Case

PR #363, opened 2023-01-28
Test cython integration in the pip_install_gmpy2.yml workflow

skirpichev:
See cython/cython#5232 for the PYTHONPATH workaround.


casevh:
Thanks,

Case

PR #366, opened 2023-02-01
Implement mpz.as_integer_ratio/to_bytes/from_bytes

### LINKED ISSUE #357: Missing methods on mpz c.f. int
The documentation claims that "It should be a drop-in replacement for Python’s long type."  Unless I misunderstood it, the mpz type must implement all public methods for the int, which is not the case for recent CPython:
```
$ python
Python 3.11.1+ (heads/3.11:ba88628808, Jan 14 2023, 08:40:06) [GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gmpy2
>>> for m in dir(int):
...     if m not in dir(gmpy2.mpz):
...         print(m)
... 
__getnewargs__
as_integer_ratio
from_bytes
to_bytes
```
Last three methods are missing.

skirpichev:
Should fix #357.  Early draft version, please don't merge unless absolutely sure.

Some benchmarks are [here](https://github.com/aleaxit/gmpy/issues/357#issuecomment-1411329795).  But note that for this size integers - arguments parsing (PyArg_ParseTupleAndKeywords branch) take ~1/3 of time.  Without this branch:
```
$ python -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'a.to_bytes(32)'
500000 loops, best of 5: 548 nsec per loop
```

### DIFF
--- a/src/gmpy2_mpz.c
+++ b/src/gmpy2_mpz.c
@@ -110,6 +110,7 @@ static PyMethodDef GMPy_MPZ_methods[] = {
     { "num_digits", GMPy_MPZ_Method_NumDigits, METH_VARARGS, GMPy_doc_mpz_method_num_digits },
     { "as_integer_ratio", GMPy_MPZ_Method_As_Integer_Ratio, METH_NOARGS, GMPy_doc_mpz_method_as_integer_ratio },
     { "to_bytes", (PyCFunction)GMPy_MPZ_Method_To_Bytes, METH_VARARGS | METH_KEYWORDS, GMPy_doc_mpz_method_to_bytes },
+    { "from_bytes", (PyCFunction)GMPy_MPZ_Method_From_Bytes, METH_VARARGS | METH_KEYWORDS | METH_CLASS, GMPy_doc_mpz_method_from_bytes },
     { NULL, NULL, 1 }
 };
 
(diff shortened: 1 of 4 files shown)


casevh:
I rarely use PyArg_ParseTupleAndKeywords due to the overhead. I'll try writing some custom argument parsing code. Even if it is only beneficial when keywords are not use, that should help the mode common use cases.

skirpichev:
> I rarely use PyArg_ParseTupleAndKeywords due to the overhead.

I will try to adapt PR, but keep in mind that the interface must be compatible with int's from/to_bytes.  It's possible to do without (see Objects/clinic/longobject.c.h and Objects/longobject.c of the CPython sources tree) with much more complex code.

skirpichev:
To use METH_FASTCALL we should drop support for older CPython versions (< 3.7).

skirpichev:
I've tested v2.1.2 with this patch on the CPython v3.6 (before conversion of int's methods to use METH_FASTCALL):
```
$ python -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'a.to_bytes(32)'
1000000 loops, best of 3: 0.878 usec per loop
$ python -m timeit -s 'from math import factorial' -s 'a = factorial(57)' 'a.to_bytes(32, "big")'
1000000 loops, best of 3: 0.798 usec per loop
```

PR #367, opened 2023-02-01
Enable printing command input lines in scripts/before*.sh

skirpichev:
An example:
https://github.com/aleaxit/gmpy/actions/runs/4060983530/jobs/6990590195
It's not clear what's was failing: curl (probably it is and --retry option may be useful) or tar.

Cleanup some debugging echo/set calls.


casevh:
Thanks.

Case

PR #370, opened 2023-02-09
Cleanup for older python versions

skirpichev:
Mostly PyIntOrLong -> PyLong renaming.


casevh:
Thanks again for the cleanup!

Case

PR #372, opened 2023-02-10
Add mpz.as_integer_ratio()

skirpichev:
trivial part of #357


casevh:
Thanks. 

It looks like we also need to add mpq.as_integer_ratio().

Case

skirpichev:
On Thu, Feb 09, 2023 at 09:23:51PM -0800, casevh wrote:
>    It looks like we also need to add mpq.as_integer_ratio().

Yes, at least these methods:
    >>> for m in dir(fractions.Fraction):
    ...     if m not in dir(gmpy2.mpq) and not m.startswith('_'):
    ...         print(m)
    ...
    as_integer_ratio
    from_decimal
    from_float
    limit_denominator

But here is some problem: Fraction's methods do support
keyword arguments, i.e. argument names are part of the API:
    >>> fractions.Fraction(1.234).limit_denominator(3)
    Fraction(4, 3)
    >>> fractions.Fraction(1.234).limit_denominator(max_denominator=3)
    Fraction(4, 3)

I'll try to convince CPython people to get rid of this (in fact,
making all arguments in the Fraction class positional-only doesn't
break own CPython tests).

Issue #371, opened 2023-02-09
Documentation has invalid function signatures


skirpichev:
Take and example:
```python
>>> import gmpy2
>>> help(gmpy2.bit_count)
Help on built-in function bit_count in module gmpy2.gmpy2:

bit_count(...)
    bit_count(x) -> int
    
    Return the number of 1-bits set in abs(x).

>>> gmpy2.bit_count(10)
2
>>> gmpy2.bit_count(x=10)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: gmpy2.gmpy2.bit_count() takes no keyword arguments
```

After [PEP 570](https://peps.python.org/pep-0570/) there is a syntax to support specification of positional-only arguments, as in the case above.  For example:
```python
>>> import inspect
>>> def foo(x):
...     pass
... 
>>> inspect.signature(foo)
<Signature (x)>
>>> foo(1)
>>> foo(x=1)
>>> foo(y=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() got an unexpected keyword argument 'y'
>>> def bar(x, /):  # here it is
...     pass
... 
>>> inspect.signature(bar)
<Signature (x, /)>
>>> bar(1)
>>> bar(x=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: bar() got some positional-only arguments passed as keyword arguments: 'x'
```

Probably, we won't support keyword arguments for most of functions (except i.e. for compatibility with int/Fraction) and function signatures must be eventually fixed (e.g. to be ` bit_count(x, /) -> int` in our example or to be `is_prime(x, n=25, /) -> bool` if positional parameter has a default value).

PR #374, opened 2023-02-10
Convert to autofunction/class's for docs/mpfr.rst

skirpichev:
* also created contexts.rst for examples/api of managing contexts
* adjust function signatures to be more compatible with the inspect module (modulo #371 issue)
* mpz/xmpz/mpq/mpfr/mpc and round2 function now have multiple signatures


skirpichev:
Ok, there are failures, but they are from the master branch...

casevh:
Thanks.

Case

PR #375, opened 2023-02-11
Add compatibility macro to work with CPython 3.12 internals too

skirpichev:
This also enables basic testing on 3.12.


casevh:
Thanks.

Case

PR #376, opened 2023-02-12
Convert to autofunction/class's for docs/contexts.rst

skirpichev:
* CTXT_Type exposed as `_context` type
* correct examples for using context managers
* minor reformatting of cython.rst
* add some missing functions to the sphinx docs


casevh:
Thanks again for all the assistance in updating the documentation.

Case

PR #378, opened 2023-02-14
Expose mpz_probab_prime_p() as a mpz method and a function

skirpichev:


casevh:
Thanks!

I was working on this but you beat me to it. I will update the docstring to match the implementation details in GMP 6. 

I'll wait a couple days for the PR to stabilize.

skirpichev:
> I will update the docstring to match the implementation details in GMP 6.

I'm not sure if we should be too verbose, i.e. reproduce the gmplib docs.

> I'll wait a couple days for the PR to stabilize.

I'll not do any changes anymore, if you are not against using METH_FASTCALL (unusual for the current codebase).  There is only tiny performance boost from this (<10%, and only for small integers).

casevh:
> I'm not sure if we should be too verbose, i.e. reproduce the gmplib docs.

I just want to state that it uses the BPSW algorithm first, then does `n-24` additional Miller-Rabin tests. (Need to check source to verify that 24 is the correct value.

> if you are not against using METH_FASTCALL

Adopting METH_FASTCALL was one of the goals for this version.

Case

skirpichev:
On Mon, Feb 13, 2023 at 11:11:11PM -0800, casevh wrote:
>    Adopting METH_FASTCALL was one of the goals for this version.

Unfortunately, handling kwargs with the METH_FASTCALL is not supported well
by the C-API CPython itself.  I think we could wait before reinventing the
wheel as the NumPy does now.

casevh:
> Unfortunately, handling kwargs with the METH_FASTCALL is not supported well
> by the C-API CPython itself.  I think we could wait before reinventing the
> wheel as the NumPy does now.

Agreed. I hoped that handling kwargs would be easier...

PR #383, opened 2023-02-16
Expand docstrings of constructors to provide more details

skirpichev:
Closes #379

PR #386, opened 2023-02-17
Correct function/methods signatures + minor fixes

skirpichev:
For constructors signatures are corrected in #383


casevh:
I hope I resolved the conflict correctly. If not, I'll fix it.

Case

skirpichev:
> I hope I resolved the conflict correctly.

No, it seems the `_printf` was restored.

PR #387, opened 2023-02-18
Improve test coverage

skirpichev:
Mostly trivial changes, but I'm not sure about `_printf`.  Maybe it's used for debugging?


casevh:
Thanks again for all the cleanup.

Yes, _printf was only used for testing the MPFR internal function. I don't think I've used it in several years.

PR #389, opened 2023-02-20
Expose context type (was _context) as context (was a function)

skirpichev:
Private _context type was exported in 1b05503 just to show context methods/attributes in docs.


casevh:
Thanks,

Case

PR #391, opened 2023-02-28
Change gcd()/lcm() to use METH_FASTCALL, optimize gcd()

skirpichev:
For random input gcd(x, y) == 1 with probability ~60%.  This commit adds a fast path in gcd for this case, where we do only arguments checking.


skirpichev:
In the master:
```
$ python -m timeit -u nsec \
>        -s 'from gmpy2 import gcd' \
>        -s 'from random import randint, seed' \
>        -s 'seed(1); a = [randint(1, 100000) for _ in range(10)]' \
>        'gcd(*a)'
200000 loops, best of 5: 1.34e+03 nsec per loop
```
On the branch:
```
$ python -m timeit -u nsec \
>        -s 'from gmpy2 import gcd' \
>        -s 'from random import randint, seed' \
>        -s 'seed(1); a = [randint(1, 100000) for _ in range(10)]' \
>        'gcd(*a)'
200000 loops, best of 5: 909 nsec per loop
```

casevh:
Thanks.

Case

PR #394, opened 2023-03-12
Add mpz.to_bytes()/from_bytes()

### LINKED ISSUE #357: Missing methods on mpz c.f. int

skirpichev:
Closes #357

Some benchmarks:
```
$ python -m timeit -s 'from math import factorial' -s 'a = factorial(57)' 'a.to_bytes(32)'
500000 loops, best of 5: 285 nsec per loop
$ python -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'a.to_bytes(32)'
500000 loops, best of 5: 486 nsec per loop
$ python -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'int(a).to_bytes(32)'
500000 loops, best of 5: 823 nsec per loop
```
It seems, int.to_bytes() is slightly better than the mpz_export.

I don't know which formatting should be used for the C code, so PEP 7 is used.  Let me know if I shouldn't use braces for single-line if/for statements or something else.  

BTW, this PR adds an optional dependency on [pytest](https://pytest.org/) and [hypothesis](https://hypothesis.readthedocs.io/), so added bulk tests check that mpz.to_bytes/from_bytes output match int's functions output for same integers.  To always keep some fixed set of tested values (e.g. for code coverage) - the "example" decorator is used.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -19,6 +19,7 @@ jobs:
           CIBW_PRERELEASE_PYTHONS: True
           CIBW_SKIP: pp*
           CIBW_BEFORE_BUILD: bash scripts/before_ci_build.sh
+          CIBW_TEST_EXTRAS: tests
           CIBW_TEST_COMMAND: python {package}/test/runtests.py
 
       - uses: actions/upload-artifact@v3
@@ -43,6 +44,7 @@ jobs:
           CIBW_BEFORE_BUILD: bash scripts/before_ci_build_apple_silicon.sh
           CIBW_SKIP: pp*
           CIBW_ARCHS_MACOS: arm64
+          CIBW_TEST_EXTRAS: tests
           CIBW_TEST_COMMAND: python {package}/test/runtests.py
 
       - uses: actions/upload-artifact@v3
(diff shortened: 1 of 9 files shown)


casevh:
Thanks.

The C formatting style is correct.

I just noticed this after I merged. Instead of using  TEMP_ALLOC(buffer,length), can we use PyBytes_FromStringAndSize(NULL, length) to create the buffer of the correct length and then manipulate the internal data directly. See also PyBytes_AS_STRING.

Case

PR #396, opened 2023-03-13
Amend f76c420 (avoid TEMP_ALLOC)

skirpichev:
a fix from the [review comment](https://github.com/aleaxit/gmpy/pull/394#issuecomment-1465314664) of #394

### DIFF
--- a/src/gmpy2_mpz_misc.c
+++ b/src/gmpy2_mpz_misc.c
@@ -1938,7 +1938,11 @@ GMPy_MPZ_Method_To_Bytes(PyObject *self, PyObject *const *args,
         blank = 0xff;
     }
 
-    TEMP_ALLOC(buffer, length);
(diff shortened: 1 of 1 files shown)


casevh:
Performance test on my machine of before and after.

Before
py311 -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'a.to_bytes(32)'
2000000 loops, best of 5: 161 nsec per loop

After
py311 -m timeit -s 'from gmpy2 import fac' -s 'a = fac(57)' 'a.to_bytes(32)'
2000000 loops, best of 5: 155 nsec per loop

So slightly faster. 

Thanks,
Case

skirpichev:
Yes, I've seen that.  Unfortunately, it still slower than the int.to_bytes().

PR #397, opened 2023-03-13
Run tests on the windows-latest too

skirpichev:
This is fast, so lets test this os too.  BTW, there are some compiler warnings.
No wheels yet.


casevh:
Based on a quick Google search, my guess is that isdigit() is expecting 'unsigned char' instead of 'char'. I'm updating my local MSYS2 environment to link with UCRT instead of MSVCRT. I probably won't be able to test until next weekend.

skirpichev:
> my guess is that isdigit() is expecting 'unsigned char' instead of 'char'.

Yes, I think so.  #398 was created to track this issue.

casevh:
Thanks,

Case

PR #406, opened 2023-04-24
Adapt conversion utils for the CPython 3.12.0a7+

### LINKED ISSUE #405: Breaking changes in Python 3.12.0a7 release
I was doing final testing before releasing gmpy2 2.2.0a1. Significant test failures.

IIRC, I last tested with 3.12.0a4 and that was successful.

I still plan to release gmpy2 2.2.0.a1 but without any wheels for 3.12.

Will update when I learn more.

skirpichev:
Fixes #405

Also do some code reorganization.


casevh:
Thanks for your quick PR.

Case

skirpichev:
@casevh, BTW CPython people suggest using to_bytes/from_bytes to get/set a PyLong, see python/cpython#101270

Issue #327, opened 2022-02-20
Creating an mpq from Decimal no longer works


oscarbenjamin:
I don't know if this was an intentional change but in gmpy2 2.0.8 the following code prints 1:
```python
import gmpy2
import decimal

print(gmpy2.mpq(decimal.Decimal(1)))
```
However in gmpy 2.1.0 ad newer it instead gives:
```console
$ python t.py 
Traceback (most recent call last):
  File "t.py", line 4, in <module>
    print(gmpy2.mpq(decimal.Decimal(1)))
TypeError: mpq() requires numeric or string argument
```
Using `mpz` here still works fine even if the decimal is not an integer e.g. `mpz(decimal.Decimal('1.5'))` gives `mpz(1)`.

I don't have any particular use for inter-operation between Decimal and Fraction so I don't have any opinion about whether it should work. I'm just reporting this as I noticed this change and I am not sure if it is intentional. It doesn't seem to be listed in the release notes:
https://gmpy2.readthedocs.io/en/latest/history.html

Decimal itself is not part of the numeric tower so I guess this change relates to gmpy2 changing to use the numeric tower.

One possibility for interacting with decimal would be the `as_integer_ratio` method:
```python
In [1]: from decimal import Decimal

In [2]: d = Decimal(3)

In [3]: d.as_integer_ratio()
Out[3]: (3, 1)
```

boegel:
This issue also surfaces in the `sympy` test suite, see https://github.com/sympy/sympy/issues/23061#issuecomment-1042462374

casevh:
It was removed during attempts to simplify the code and decrease the overhead of argument type checking. Unfortunately, it was not well documented.

The string conversion routines for creating an 'mpq' or 'mpfr' support exponents so the following does work:

```
>>> mpq(str(Decimal("1.2e100")))
mpq(12000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000,1)
>>> mpfr(str(Decimal("1.2e100")))
mpfr('1.2000000000000001e+100')
```
but it doesn't work for `mpz`

```
>>> mpz(str(Decimal("1.2e100")))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid digits
```

There is special logic in the `mpq` and `mpfr` string conversion routines that handles exponents properly but it doesn't exist in the `mpz` conversion routine.

[...]

PR #413, opened 2023-05-30
Misc fixes

skirpichev:
* re-add support for Decimal->mpq conversion, closes https://github.com/aleaxit/gmpy/issues/327
* bulk tests for int(mpz(int(n)))==int(n) invariant


casevh:
Shouldn't `res` be decremented after calling `GMPy_MPQ_NewInit` ?

skirpichev:
> Shouldn't res be decremented after calling GMPy_MPQ_NewInit ?

Indeed.  First commit was updated to address this.

casevh:
Thanks.

Case

Issue #414, opened 2023-06-10
Incompatible (wrt the MPC library) parsing in the master


skirpichev:
After 43e8de6, the `GMPy_RemoveIgnoredASCII()` strips white spaces as well as underscores.  This breaks MPC-compatible string input parsing, e.g.:
```python
>>> gmpy2.mpc('(1 2)')
mpc('12.0+0.0j')
```
Not sure if it worth to support this feature forever (i.e. maybe we could deprecate this to simplify parsing).  But this behaviour was documented in the sphinx docs before, the mpc.rst of the last stable release (2.1.5) says "In addition to the standard Python string representation of a complex number: "1+2j", the string representation used by the MPC library: "(1 2)" is also supported".

casevh:
I think we should support the MPC format. Something like the following should work:

1. Convert input string to ASCII.
2. If the first character is '(' and the last character is ')' then remove underscores only and call ` mpc_strtoc`.
3. Otherwise, convert the input string using `GMPy_RemoveIgnoredASCII()` and proceed with the current code.

PR #415, opened 2023-06-16
Misc fixes

skirpichev:
* expand mpfr() docstring: binary exponents
* expand mpq() docstring: base=0 case
* expand mpfr() docstring: mention float() and float.fromhex()
* restore MPC-compatible string parsing, closes https://github.com/aleaxit/gmpy/issues/414


casevh:
Thanks again.

Case

PR #432, opened 2023-09-15
pyproject.toml: Update [build-system] requires

mkoeppe:
Removing `pip` and `wheel`, which are not needed.
(for wheel, see https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use)


casevh:
Thanks.

I think the failure in the tests is temporary. I will watch it.

PR #435, opened 2023-10-01
Skip mpmath test if module not available

tornaria:
Since gmpy2 is a build dependency for mpmath, this allows testing gmpy2 before mpmath is built.

### DIFF
--- a/test/test_mpfr.py
+++ b/test/test_mpfr.py
@@ -3,7 +3,6 @@ from decimal import Decimal
 import pytest
 from hypothesis import given, example, settings
 from hypothesis.strategies import floats
-import mpmath
 
 import gmpy2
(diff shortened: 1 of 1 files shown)


casevh:
Thanks!

skirpichev:
> Since gmpy2 is a build dependency for mpmath

Actually, it's not.  You can test the mpmath without gmpy2.

tornaria:
> > Since gmpy2 is a build dependency for mpmath
> 
> Actually, it's not. You can test the mpmath without gmpy2.

I stand corrected. Neither build or runtime dependency.

However, I believe mpmath should be faster with gmpy2, so we ship mpmath with gmpy2 as a runtime dep. It seems better to build gmpy2 first and skip one test, then build mpmath and test with gmpy2. In any case, this PR allows one to build and test both packages independently of each other.

Issue #377, opened 2023-02-14
Use pytest for testing?


skirpichev:
The doctest module does actually check the string output, not the returned value or the exception raised.  So, it's might be a good idea to switch tests (& test_cython, perhaps, like [this](https://shwina.github.io/cython-testing/)) to some more common testing framework.

The unittest module seems to be too verbose, so I'll suggest the pytest (see the mpmath as an example (https://github.com/mpmath/mpmath/tree/master/mpmath/tests).  It has many plugins, ex. an integration with the [hypothesis](https://hypothesis.readthedocs.io/en/latest/) library, useful for testing specifications and invariants.

UPD: Also, it's not documented how to test installed package (or run tests in the source tree).

casevh:
I think that would be great. In the past, I tried to minimize the need for external libraries when compiling gmpy2. But the development world has changed.

Case

skirpichev:
> I tried to minimize the need for external libraries when compiling gmpy2

pytest will be a requirement only for testing.

casevh:
> pytest will be a requirement only for testing.

Understood.

casevh:
I like the pytest format.

I've started to add MPFR 4.2.x specific functions. What is the best way to handle version dependent tests? 

I've thought of the following options:

1) A version specific file with the version check done in runtests.py. I don't like adding complications to runtests.py.

2) Place them in the same file (test_mpfr.py) and surround the tests with a version check. This keeps runtests.py simple but can make test_mpfr.py larger and (possibly) more difficult to understand.

3) Create a file called test_mpfr_v420.py that contains tests for functions added in MPFR 4.2.0 and add a single version check at the beginning of the file.

I prefer option 3 but I'm open to any other options.

skirpichev:
That depends on the number of tests.  You can use skip or [skipif](https://docs.pytest.org/en/latest/reference/reference.html#pytest-mark-skipif-ref) decorators to exclude either a single test function or an entire test file. Separate test file, probably, does make sense if there are many tested functions and/or if we want keep that file forever.  But I would expect that at some point new functions will be mandatory, so option 2 sounds better (or we can split test_mpfr.py to several pieces based on functionality tested, e.g. arithmetics, elementary functions, etc).

kloczek:
Thank you 😄 👍 
Do you have any plans to release final 2.2.0 soon? 🤔 
(just asking 😋 )

PR #437, opened 2023-10-10
Port some tests to pytest framework (step 2)

skirpichev:
see https://github.com/aleaxit/gmpy/issues/377

a continuation of #408


casevh:
Thanks.

PR #440, opened 2023-10-18
Add __version__ module attribute

skirpichev:
This is a more standard way to embed version info.  Recognized, e.g. by pytest.importorskip.  

Also, simplify version handling in docs/conf.py.

Probably, we could also deprecate gmpy2.version() function.  @casevh, what do you think?


casevh:
Thanks for adding __version__. I think we should keep gmpy2.verison(). I don't think it is worth the disruption.

PR #449, opened 2023-10-30
Misc fixes

skirpichev:
9fce264 Don't use private API: _Py_HashPointer
1de1bc2 Add workaround for shutil.copytree on CPython 3.7
2b805ac Run Cython tests in coverage job
b723d71 Run _set_default_context() fixture per function
ac2cfe2 Add pytest_report_header(), like test/runtests.py


casevh:
Thanks again for all the cleanup work.

PR #452, opened 2023-11-04
Misc fixes

skirpichev:
~~2e96fb6 Avoid computing _PyHASH_MODULUS in GMPy_MPQ_Hash_Slot()~~
~~324e284 Fix compiler warning~~
8a7b1ff Fix rst markup in docstrings (for emphasis)
12afc77 Use ``gmpy2.__package__`` to get project name in sphinx
b470a3b Set doctest_global_cleanup sphinx directive

PR #453, opened 2023-11-06
Fix compiler warning in _mpfr_hash()

skirpichev:
Also drop some inaccessible code in GMPy_MPC_Hash_Slot()

-----

This is cherry-picked from #452.  Old discussion: https://github.com/aleaxit/gmpy/commit/324e284037a334bf9523ffcf508efe3a854706c4#commitcomment-131773663

### DIFF
--- a/src/gmpy2_hash.c
+++ b/src/gmpy2_hash.c
@@ -117,7 +117,9 @@ _mpfr_hash(mpfr_t f)
         }
         else {
 #if PY_VERSION_HEX >= 0x030A00A0
-            return PyBaseObject_Type.tp_hash(f);
+            /* The default object hash implementation in the CPython
+             * accepts void* pointer. */
+            return PyBaseObject_Type.tp_hash((PyObject*)f);
 #else
             return _PyHASH_NAN;
 #endif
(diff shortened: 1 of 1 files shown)


skirpichev:
I think the only issue with using ``PyBaseObject_Type.tp_hash`` slot is that we can't guarantee that in future CPython versions it's argument will be interpreted as ``void *``.  But I think it's highly unlikely to be changed.  We can ask to document this.

casevh:
Oops. Sorry about the confusion. I missed that.

Again, sorry about my confusion.

skirpichev:
> Oops. Sorry about the confusion. I missed that.

@casevh, actually I didn't expect this PR will be merged immediately.  Rather it was factored out to discuss hashing of NaN's.

Here is the old discussion (keep in mind next time it's better to comment code in pr, e.g. "File changed" tab, not individual commits: this will be more transparent to other readers): https://github.com/aleaxit/gmpy/commit/324e284037a334bf9523ffcf508efe3a854706c4#commitcomment-131773663
and my above comment: https://github.com/aleaxit/gmpy/pull/453#issuecomment-1794117906

So, using PyBaseObject_Type.tp_hash is an exact equivalent of ``_Py_HashPointer()``.  While it's not documented, it's hard to imagine a default implementation of the object type hashing to be different in the interface (i.e. dereference the pointer).

I think we have these options:
1. Ask CPython devs to document hashing of the object type (like for ``id()`` builtin).
2. Ask them to restore ``_Py_HashPointer()`` as a public interface (like https://github.com/python/cpython/issues/111545)

(1) looks to be better for me, it's hard to imagine where new public interface could be used.  Only for hashing of NaN's: https://github.com/python/cpython/blob/ba8aa1fd3735aa3dd13a36ad8f059a422d25ff37/Doc/library/stdtypes.rst?plain=1#L789-L790

casevh:
I agree that option (1) is the best choice and that any changes are unlikely.

skirpichev:
Ok, I've asked this in python/cpython#111389.

PR #460, opened 2023-12-04
Deprecate local_context() and update contexts.rst

### LINKED ISSUE #454: Deprecate local_context()?
Ordinary context constructors could be used instead, isn't?

For example:
```pycon
>>> with context() as ctx:
...     ctx.allow_complex = True
...     print(sqrt(-2))
... 
0.0+1.4142135623730951j
>>> ctx = get_context()
>>> ctx.allow_complex
False
>>> print(sqrt(-2))
nan
```

skirpichev:
* use context() instead of local_context() in the tutorial
* drop "Comments on the ieee context" section
* drop "Nested contexts and coding style"
* deprecate local_context() and update context() docstring, closes #454
* drop CTXT_Manager_Type
* update 1st paragraph of docs/contexts.rst
* change local_context() -> context() in tests and docs
* slightly expand tutorial.rst and drop "Contexts and the with statement" section
* improve test coverage for GMPy_CTXT_Context()


gmpy2/__init__.py
casevh, line 1: local_context has a different behavior when args == (). The replacement function needs to support that.

I'll implement the replacement in C and expose it as _local_context. Then this helper function can be changed to

return _local_context(*args, **kwargs)

casevh: I've added _local_context(). The tests pass when I use modify __init__.py to import _local_context as local_context.

casevh:
Thanks for the cleanup. I'll merge it later this evening.

gmpy2/__init__.py
skirpichev: Indeed, I forgot the ``context()`` has a different semantics wrt defaults.  On another hand, maybe we *want* to change this?  I.e. if ``cxt`` is not specified, shouldn't context() be based on the current context, not on the default?  This will be an incompatible change, of course.

Meanwhile, I did ``local_context()`` in C, based on your function.

skirpichev:
Hmm, there are some uncovered lines.  I'll add more tests.

skirpichev:
Ok, there is no regressions in tests.

gmpy2/__init__.py
casevh: > Indeed, I forgot the `context()` has a different semantics wrt defaults. On another hand, maybe we _want_ to change this? I.e. if `cxt` is not specified, shouldn't context() be based on the current context, not on the default? This will be an incompatible change, of course.
> 

IIRC, `context()` was originally called `new_context()`. From that history, it makes sense that `context()` creates a new default context. And that's why `context()` and `local_context()`. 

I'm interested in allowing `context()` to use gmpy2.DefaultContext to get a dictionary that defines the default context. Then add context.as_dict to create such a dictionary. Lots of subtle details that shouldn't be part of this PR.

skirpichev: > And that's why context() and local_context().

Well, I think basically there are two distinct cases where ``context()`` could be used.

First, "reset to defaults":
```python
ctx = get_context()
ctx.precision += 10
ctx.allow_complex = True
<lots of other customization and computations>
# now reset to defaults:
set_context(context())
...
```

Second, as a context manager (in "with" statement):
```python
<some computation>
with context() as ctx:
    # temporary adjust few options, e.g.
    ctx.precision += 100
    <do computation> 
<continue with old options>
```

In the second case we, probably, want the ``local_context()`` semantics.  But, obviously, this will break first scenario.  To overcome this we could add an instance method to reset context.

casevh: Or we accept there are two use cases and keep both `context()` and `local_context()` and remove the warning.

What about writing the first use case as:

```
gmpy2.DefaultContext = get_context().as_dict()
<<do stuff>>
set_context(context())
```

where context() become context(**gmpy2.DefaultContext). The user can control the default context. By using a dictionary, custom contexts can be easily distributed. This a safe change since the default DefaultContext is just the one we are currently using so there is no difference unless DefaultContext is intentionally changed.

skirpichev: > Or we accept there are two use cases and keep both context() and local_context() and remove the warning.

Maybe.  But I don't sure we have good naming for this stuff, esp. ``local_context()``.

> The user can control the default context.

Users can control the default context right now, e.g.:
```python
# at the beginning of the session/program:
from gmpy2 import <...>
default_context = get_context().copy()
<do stuff>
set_context(default_context.copy())
```

> By using a dictionary, custom contexts can be easily distributed.

That's a good point.  We also could make contexts pickleable.

PR #466, opened 2024-02-03
Misc fixes

skirpichev:
* drop workaround for CPython 3.13, see python/cpython#112115
* port more functions/methods to METH_FASTCALL convention
* simplify _mpmath_normalize/_mpmath_create
* update copyright


casevh:
Thanks for the cleanup.

Issue #467, opened 2024-02-20
Alternatives to using private API for conversion to/from CPython int's


skirpichev:
Currently, we are using a bunch of private CPython functions to provide fast mpz <-> int conversion: see [GMPy_MPZ_From_PyLong](https://github.com/aleaxit/gmpy/blob/141eb88c9c4ab810d7988f0ebd17c83a6c11a8b7/src/gmpy2_convert_gmp.c#L42), [GMPy_PyLong_From_MPZ](https://github.com/aleaxit/gmpy/blob/141eb88c9c4ab810d7988f0ebd17c83a6c11a8b7/src/gmpy2_convert_gmp.c#L149C1-L149C21) and 
https://github.com/aleaxit/gmpy/blob/141eb88c9c4ab810d7988f0ebd17c83a6c11a8b7/src/gmpy2_convert.h#L135-L155
There should be a better way!

I see several variants to deal with this problem.

1. The CPython 3.13 got [PyLong_AsNativeBytes()](https://docs.python.org/3.13/c-api/long.html#c.PyLong_AsNativeBytes) and [PyLong_FromNativeBytes()](https://docs.python.org/3.13/c-api/long.html#c.PyLong_FromNativeBytes) function to import/export arbitrary int's.  We can use these functions and handle small integers case separately with [PyLong_AsLong()](https://docs.python.org/3.13/c-api/long.html#c.PyLong_AsLong)/[PyLong_FromLong()](https://docs.python.org/3.13/c-api/long.html#c.PyLong_FromLong).  In general, this should be much slower than the current solution.  On another hand, I think that for real world scenarios - small integers case should be most important.

2. There is a [proposal](https://github.com/python/cpython/issues/102471#issuecomment-1620352561) to add ``mpz_import/export``-like functions for PyLong's C API, but it seems that CPython core developers aren't very enthusiastic on this idea (see https://github.com/python/cpython/issues/111140#issuecomment-1928654497).

3. Another possibility is the ``mpz_limbs_write()``-like API to access unsigned bigint as an array of "digits", see [this](https://github.com/python/cpython/issues/111140#issuecomment-1935356765).  With that kind of API on the CPython side - the gmpy2 could do fast conversion to/from int's using mpz_import/export functions.

I'm planning to open a discussion thread on the d.p.o, that will propose the 3rd option, but first I would appreciate any feedback here.  CC @oscarbenjamin: perhaps this does make sense for the python-flint.

skirpichev:
[...]

Issue #352, opened 2022-12-20
Building wheels with GMP et al


oscarbenjamin:
If this is a bad place to open this discussion then let me know and I'll take it elsewhere.

I've been working on python-flint which is another project that packages similar libraries to gmpy. Along with GMP and MPFR, python-flint also includes Flint and Arb so it is similar in some ways to gmpy but with a different overall scope. The main thing I've been working on is producing wheels (https://github.com/fredrik-johansson/python-flint/issues/1) and I'm pretty close to having that working with some of that work inspired by the way that gmpy does things.

I wonder if there is scope for collaborating and learning from each other about exactly what is a good way to build something like this because it seems like a very much uphill struggle with current Python packaging infrastructure (although things are getting better over time).

In particular I have questions right now about how exactly gmpy makes wheels for Apple arm64. I see that recent commits include a patch to GMP for this platform. Is that necessary? Does something not work without it?

Has the gmpy project considered moving to meson or cmake etc because I see some suggestion that other scientific Python projects are moving that way. Could there be any scope to work together on moving towards new build systems?

casevh:
Hi Oscar,

This is a fine forum for these discussions.

I skimmed your work that you've done on python-flint. I think you've encountered most of the same issues with complex builds.

Regarding Apple arm64 builds - yes, the patch is required. Apple reserved a register for their use. GMP 6.2.1 uses that register. The patch modifies GMP 6.2.1 to save/restore that register. (https://gmplib.org/list-archives/gmp-discuss/2022-July/006821.html) Without the patch, it will crash frequently. (See https://github.com/aleaxit/gmpy/issues/350)

Some comments on Windows

Compiling gmpy for Windows has always been a challenge. -cmingw32 worked well for 32-bit versions of Python but required extensive hacking to use mingw64 compilers on 64-bit Windows. (I originally used MPIR and MSVC but MSVC didn't support --enable-fat. I didn't want to compile for the lowest common CPU type nor try to release CPU specific versions.)

I compile the Windows binaries locally. I just extracted a copy of the cygwincompiler.py file from numpy and replace the file that comes with Python with a slightly modified copy. I don't compile anything else on Windows so it works for me. It would be great if the numpy version of -cmingw32 could be added to setuptools as -cmingw64. That would eliminate any risk of breakage for existing code that patches -cmingw32 on the fly.

[...]

PR #474, opened 2024-03-23
Drop gmpy2 directory & move source files to src/gmpy2/

skirpichev:
This uses more modern directory layout (like python-flint does, for example) and avoids extra pure-python code while doing import.  The drawback is that you have to use extra ``--follow`` arguments for some git commands (e.g. ``git log``) to access early history.

With this patch:
```
$ time python -c 'from gmpy2 import *'

real    0m0.065s
user    0m0.050s
sys     0m0.015s
```
On the master:
```
$ time python -c 'from gmpy2 import *'

real    0m0.120s
user    0m0.095s
sys     0m0.024s
```

Uses patched delocate from my repo: git+https://github.com/skirpichev/delocate.git@fix-lib-sdir


skirpichev:
@casevh, this is a draft pr, but let me know if you like this idea.

casevh:
I installed a Linux wheel. It imports and runs just fine. I then looked at the file layout. The shared library is installed in `site-packages/`, the include files are installed in `site-packages/gmpy2/`, and the supporting libraries are installed in `site-packages.gmpy2.libs`.

The following comments are based on memories from a discussion that occurred several years. `gmpy` and `gmpy2` always used to install its shared library in `site-packages`. When the `gmpy2` direcctory was added as part of support for `setuptools`, the shared library was moved to the `gmpy2` directory and loaded via `gmpy2/__init__.py`. I questioned the rationale. I was left with the impression that using `gmpy2/__init__.py` was the correct approach and that Python extensions shouldn't just be dumped into `site-packages`.

I personally liked having `gmpy2...so` in `site-packages`. (For many years, the Windows binaries where distributed as a single statically-linked DLL that was placed in `site-packages`. This predated pip.) But I don't want to rely on an unsupported file location that might depracted in the future.

I do like the faster import time (so +1), but I don't know if it is worth the possible long-term stability risks (-0.5) and the code churn (also -0.5).

I'll continue to think about this...

skirpichev:
> But I don't want to rely on an unsupported file location that might depracted in the future.

I don't sure I realize which scenario do you mean.   The module so-file in the site-packages directory is a long-standing default for the ``setuptools.Extension()``.

skirpichev:
This is ready for review.

skirpichev:
Hmm, it seems delvewheel in this settings puts all libraries to the site-packages/ dir.  Will think if it's possible to workaround this.

Meanwhile, mark this as a draft.  Please don't merge.

skirpichev:
Well, the problem is that PE format has nothing like RPATH in ELF...  So, options are either
1. put dll's alongside with the module dll.
2. use ``__init__.py`` file and put here suitable os.add_dll_directory() calls, that will point to the ``gmpy2.libs`` dir.

The delvewheel uses both options, respectively, (1) - if extension module is top-level, and (2) - if it has ``__init__.py`` file.

I think (1) is fine, if we will also use name-mangling on M$ Windows (now - turned off).  Not sure if it's file without name-mangling.  Maybe the better alternative is postpone this pr and, instead, package GMP, MPFR and MPC libraries as separate packages like scipy-openblas64?

casevh:
I propose option (1) but renaming the DLLs to well-defined names. The
renaming is done before gmpy2 is compiled.

Look at https://github.com/emilbayes/rename-dll for a utility to do the
renaming.

For a naming convention, how about libgmp-10.dll becomes
lib_gmpy2v2_2-10.dll and similarly for mpfr and mpc. The implies that the
DLLs were distributed as part of gmpy2 version 2.2.x. The name would change
if the DLLs change in a backwards incompatible way. And that should trigger
a new version of gmpy2.

We won't conflict with standard names. And C extensions are able to use the
renamed DLLs.

skirpichev:
> For a naming convention, how about libgmp-10.dll becomes lib_gmpy2v2_2-10.dll and similarly for mpfr and mpc.

I guess you meant something like "libgmp_gmpy2v2_2-10.dll".  Your variant doesn't include library name.

> [rename-dll](https://github.com/emilbayes/rename-dll) for a utility to do the renaming.

Do you suggest to restrict this workaround only for M$ Windows?  I would prefer a common setup for all platforms.  That's why I'm thinking about separate packages for libraries.  Probably, this should be discussed in #352: now we have binary wheels for all platforms, what are your plans for the referenced issue?

casevh:
> 
> I guess you meant something like "libgmp_gmpy2v2_2-10.dll". Your variant doesn't include library name.
> 
Correct.
>
> Do you suggest to restrict this workaround only for M$ Windows? I would prefer a common setup for all platforms. That's why I'm thinking about separate packages for libraries. Probably, this should be discussed in #352: now we have binary wheels for all platforms, what are your plans for the referenced issue?

It looks like flint has updated their build system to meson. https://github.com/flintlib/python-flint/issues/52

I would like to focus on resolving the challenges with creating extensions that want to use gmpy2's C-API. I think it is fine to require installing a binary wheel of gmpy2 to get the GMP, MPFR, and MPC libraries. The fundamental issues are (1) predictable names and (2) finding the required files. 

For Windows, I slightly prefer to place the supporting files in the same directoy as the gmpy2 pyd file. That layout has been used in the past. This does require importing gmpy2 before importing the C extension but I don't think that is unreasonable.

casevh:
There are many unresolved questions in this PR. Would you be okay if we release gmpy2 2.2.0 quickly to address support for Python 3.12 and 3.13 and delay this PR until gmpy2 2.3 (hopefully released when Python 3.14 is released)?

If you agree, I'll bump the version to b1 and release it ASAP for testing. Then release rc1 shortly after 3.13.0b1 is released.

skirpichev:
@casevh, sure!  It's a draft pr, it shouldn't block anything.

But maybe you would like also to do a minor release for 2.1.x with support for CPython 3.12/13?

casevh:
> But maybe you would like also to do a minor release for 2.1.x with support for CPython 3.12/13?
@skirpichev A source-only release for Linux distributions would be easy. I need to see if I can replicate the 2.1.x process for compiling the binary wheels for Windows. 

Maybe just release 2.1.6 as source-only and release 2.2.0 as rc1. The "pip install gmpy2" should grab the rc1 version and a standard 2.1.6 version is available for Linux distributions.

 I'll probably increment the version to b1 this evening. Are there any last minute changes you want to make?

skirpichev:
> Maybe just release 2.1.6 as source-only

Does make sense for me.

> Are there any last minute changes you want to make?

No:)  I would like to solve #467, but it seems that new C API will not land in the CPython 3.13.

Issue #475, opened 2024-04-20
Please make a non-preview release with Python 3.12 support — thank you!


hartwork:
Hello gmpy! :wave: 

Linux distros are already patching Python 3.12 support into 2.1.5, e.g. Gentoo and Void apply patch https://raw.githubusercontent.com/tornaria/void-packages/722b32aa405804b79a74256708de6a511e255b4b/srcpkgs/python3-gmpy2/patches/cleanup-object-caching.patch downstream to unblock Python 3.12. `pip install gmpy2` remains broken for Python 3.12 however, and as a result Python 3.12 support of reverse dependencies like https://github.com/badkeys/badkeys is broken when installed through pip, directly. Is there a chance for a non-preview release with Python 3.12 support? Even a backport to 2.1.x a la 2.1.6 could improve the situation, if feasible. What do you think?

Best, Sebastian

CC @hannob

iyanmv:
Arch Linux has Python 3.12 now in testing repositories, I guess in some days it will arrive to the stable repos. [They have patched the package providing gmpy2](https://gitlab.archlinux.org/archlinux/packaging/packages/python-gmpy2/-/commit/1c0d2d08d15b7b3bea012df2e91dbe0166be0afc) to work with Python 3.12, but this will be an issue for users creating virtual envs that require gmpy2.

casevh:
I have created a branch for 2.1.6 with the necessary patches for Python 3.12 and 3.13. It will only be a source-only release. I want to release 2.2.0rc1 with binary wheels at the same time.

skirpichev:
The gmpy2-2.1.6 also has CI build failures.  I think, it will be not so difficult to solve: we just need to update build images and include/exclude some CPython versions from the build_wheels.yml.  Does it make sense for you, @casevh?  Or not worth it?

casevh:
It would take me some time to recreate the environment that I used for the Windows builds. It used mingw64 for the entire build process. That required hacking of distutils. I would rather focus on 2.2.0.

skirpichev:
Fix for gmpy2-2.1.6 branch: https://github.com/aleaxit/gmpy/pull/479

skirpichev:
@casevh, let me know if I can help more with new releases.

hannob:
Any progress here?

[...]

PR #483, opened 2024-05-20
Fix CI on CPython 3.13 due to difference _decimal/_pydecimal 

skirpichev:


casevh:
Thanks for fixing these issues. It's getting hard to keep up with all the changes just required to keep building wheels.

skirpichev:
@casevh, please let me know if something else blocks first beta release or #475.

casevh:
@skirpichev I'm only aware of one minor issue with 3.13. Here is an example:

>>> int('11') is int('11')
True
>>> int(mpz('11')) is int(mpz('11'))
False

I think the fix is in GMPy_PyLong_From_MPZ. It should first check if an mpz can be converted to a C long or ssize_t and use the appropriate PyLong_FromLong() or PyLong_FromSsize_t(). I haven't check the CPython source code, but this should allow interning to work and support creation of compact values.

I could do a beta release without this fix. With this fix I would jump to rc1 status. Then I would do a joint release of 2.1.6 source and 2.2.0rc1 with source and wheels.

BTW, the fix for CPU detection resulted in a 10 to 15% improvement in multiplication performance.

skirpichev:
> I'm only aware of one minor issue with 3.13.

Hmm, this is broken on 3.12 and 3.11 as well.

> I could do a beta release without this fix.

No, I think we should fix this.  I will take a look today.

PR #486, opened 2024-06-03
Use PyLong_GetSign()

skirpichev:
See https://github.com/python/cpython/pull/116561 and https://github.com/capi-workgroup/decisions/issues/19.  I'm not happy with this API and it seems that discussion is ongoing.  Maybe it's wise to wait some time...

### DIFF
--- a/src/gmpy2_convert.h
+++ b/src/gmpy2_convert.h
@@ -149,7 +149,7 @@ extern "C" {
 #  define GET_OB_DIGIT(obj) ((PyLongObject*)obj)->long_value.ob_digit
 #  define _PyLong_DigitCount(obj) (((PyLongObject*)obj)->long_value.lv_tag >> 3)
 #else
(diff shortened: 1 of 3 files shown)


skirpichev:
Probably, discussion in https://github.com/capi-workgroup/decisions/issues/19 is ended.

casevh:
Should this be merged now and then released as 2.2.0 or should this wait until 2.2.1?

skirpichev:
> Should this be merged now and then released as 2.2.0

I don't think so.  https://github.com/capi-workgroup/decisions/issues/29 is open.  We should wait, see also https://discuss.python.org/t/c-api-working-group-vacancy-again/56167

This pr shouldn't block 2.2 release.  I'll just mark it as a draft.

skirpichev:
I don't like it, but probably https://github.com/capi-workgroup/decisions/issues/35 API will have ``negative`` field.  So, probably we can skip this.

PR #487, opened 2024-06-27
Use codecov-action v4

skirpichev:
It seems, https://app.codecov.io/gh/aleaxit/gmpy - not updating. And coverage for new pull requests is missing too:
https://app.codecov.io/gh/aleaxit/gmpy/pulls?prStates%5B0%5D=OPEN

This update codecov-action version and add fail_ci_if_error option to prevent CI failures on the master.  In this form, submitting data works in mpmath's prs.

If someone menages to add CODECOV_TOKEN (as it's explained in https://github.com/codecov/codecov-action/?tab=readme-ov-file#usage)
- then we could remove fail_ci_if_error option (defaults to true).


casevh:
Thanks.

PR #495, opened 2024-07-15
Use PEP 757 C-API to import/export ints 

### LINKED ISSUE #467: Alternatives to using private API for conversion to/from CPython int's

skirpichev:
Closes #467

### DIFF
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -177,6 +177,14 @@ static PyObject *GMPyExc_Overflow = NULL;
 static PyObject *GMPyExc_Underflow = NULL;
 static PyObject *GMPyExc_Erange = NULL;
 
+/*
+ * Parameters of Python’s internal representation of integers.
+ */
+
+
+size_t int_digit_size, int_nails, int_bits_per_digit;
+int int_digits_order, int_endianness;
+
 
 /* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
  * End of global data declarations.                                        *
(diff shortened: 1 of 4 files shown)


skirpichev:
@casevh, at this point I would appreciate your feedback on https://github.com/capi-workgroup/decisions/issues/35.  Proposed API has noticeable speed regression for int->mpz conversion.  I'm not sure if this is acceptable for gmpy2.

skirpichev:
@casevh, this is ready for review

casevh:
Thanks for all the work on this.

PR #502, opened 2024-07-31
Extend fallthrough safety to LLVM in gmpy2_mpmath.c

r-barnes:
The current fallthrough comments only suppress warnings for GCC. `__attribute__((fallthrough));` works for both GCC (>=7) and LLVM. C23 will introduce `[[fallthrough]]`.

### DIFF
--- a/src/gmpy2_macros.h
+++ b/src/gmpy2_macros.h
@@ -726,3 +726,14 @@ GMPy_Context_##NAME(PyObject *self, PyObject *args) \
     } \
     return GMPy_Number_##NAME(PyTuple_GET_ITEM(args, 0), PyTuple_GET_ITEM(args, 1), context); \
 }
+
+/*********************************************************************/
+
+#if defined(__has_attribute)
+#  if __has_attribute(fallthrough)
+#    define GMPY_FALLTHROUGH __attribute__((fallthrough))
+#  endif
+#endif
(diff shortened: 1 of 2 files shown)


skirpichev (commented):
This seems to be gcc/clang-specific.

Not a surprise, this will break CI on Windows:
```
    D:\a\gmpy\gmpy\src\gmpy2_mpmath.c(290): warning C4013: '__attribute__' undefined; assuming extern returning int
    D:\a\gmpy\gmpy\src\gmpy2_mpmath.c(290): error C2065: 'fallthrough': undeclared identifier
    D:\a\gmpy\gmpy\src\gmpy2_mpmath.c(297): error C2065: 'fallthrough': undeclared identifier
    D:\a\gmpy\gmpy\src\gmpy2_mpmath.c(300): error C2065: 'fallthrough': undeclared identifier
```

skirpichev (commented):
Sorry, I still don't see value in this pr.  Commentary is recognized by gcc and clang.  It's harmless and useless for M$ VC (BTW, does it has ``-Wimplicit-fallthrough`` analog?).

skirpichev:
> clang does not recognize comments as having semantics

Ah, then it makes some sense.

> so the code as-is will fail to compile using clang with -Wimplicit-fallthrough enabled.

(only with -Werror)

skirpichev:
Now this has merge conflicts.

PR #513, opened 2024-09-13
Add signal handler to catch SIGFPE on the GMP 6.3+

### LINKED ISSUE #497: Probably incorrect error message.
I tried (unsuccessfully) to find a next term for https://oeis.org/A348651

Using GMP:
```
#include <stdio.h>
#include <stdint.h>
#include <gmp.h>

#define _FAC_13 6227020800

void f1 (uint64_t x)
{
        mpz_t tmp;

        mpz_init (tmp);
        mpz_fac_ui (tmp, x);
        printf ("%ld\n", mpz_popcount(tmp));
}

int main (void)
{
        f1(_FAC_13);
};
```

On an instance with 300G of RAM, it consumed up to ~80-100G and stopped with the error: "gmp: overflow in mpz type".
Ouch!

Also I tried gmpy:
```
from gmpy2 import fac, popcount

def A348651(n):
    return popcount(fac(fac(n)))

print (A348651(13))
```

It also consumed that much RAM and stopped with error "Floating point exception".

Maybe you should know about such behaviour.

skirpichev:
Since 6.3.0 some memory errors in mpz/realloc.c trigger SIGFPE exception.  I think it's an upstream bug: helpful message should be emitted instead.

This patch install SIGFPE handler, that does this.

Closes #497

### DIFF
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -96,6 +96,7 @@
 #include <math.h>
 #include <float.h>
 #include <ctype.h>
+#include <signal.h>
 
 /*
  * we do have a dependence on Python's internals, specifically:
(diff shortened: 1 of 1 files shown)


casevh:
> GMP_ERROR_DIVISION_BY_ZERO and GMP_ERROR_SQRT_OF_NEGATIVE, in principle, shouldn't be triggered by some code in the gmpy2 codebase. Maybe I should check only GMP_ERROR_MPZ_OVERFLOW?

While GMP_ERROR_DIVISION_BY_ZERO and GMP_ERROR_SQRT_OF_NEGATIVE shouldn't occur, I think it's fine to catch all the errors.

PR #516, opened 2024-09-13
Fix memory leak in GMPy_CTXT_Exit()

### LINKED ISSUE #515: Memory leak when using gmpy2.ieee context manager
When using the ieee context manager repeatedly, memory is leaked using gmpy latest master branch (c7629dd). The problem can be reproduced in the following way:

```
$ ulimit -v $((1024*256))
$ cat test.py
import gmpy2

while True:
    with gmpy2.ieee(64):
        x = 1
$ python test.py
Traceback (most recent call last):
  File "test.py", line 4, in <module>
MemoryError
```

The test case works fine (i.e. runs forever) when using gmpy 2.1.5. Thanks for looking into this!

skirpichev:
Closes #515

### DIFF
--- a/src/gmpy2_context.c
+++ b/src/gmpy2_context.c
@@ -184,7 +184,9 @@ GMPy_CTXT_Enter(PyObject *self, PyObject *args)
 static PyObject *
 GMPy_CTXT_Exit(PyObject *self, PyObject *args)
 {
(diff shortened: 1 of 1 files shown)


casevh:
Thanks for the quick fix.

PR #519, opened 2024-09-25
Misc fixes

skirpichev:
* more samples + test str(int)==str(mpz) in test_mpz_conversion_bulk()
* more correct error message for MPZ_OVERFLOW error in GMP

### DIFF
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -579,7 +579,7 @@ gmp_abort_handler(int i)
             printf("gmp: root of negative\n");
         }
         else {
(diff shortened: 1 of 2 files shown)


casevh:
`gmpy` used the custom memory allocation feature of GMP to to wrap the malloc calls and raise `MemoryError` if they failed. It didn't catch all core dumps but this patch looks like a better option. Is there any advantage to raising `MemoryError` in a future release? (Just having more informative error messages on a core dump is a definite improvement.)

PR #521, opened 2024-10-13
Misc fixes

### LINKED ISSUE #504: Incorrect signature for gmpy2.digits() in docs
E.g.:
```pycon
>>> from gmpy2 import digits
>>> help(digits)
Help on built-in function digits in module gmpy2.gmpy2:

digits(...)
    digits(x, base=10, prec=0, /) -> str

    Return string representing a number x.

```

In fact, output is different: it's a string for mpz/mpq, but a 3-tuple for mpfr and a 2-tuple for mpc.  Signature looks to complex to be described in details, probably ``-> str | tuple`` will be enough.  Documentation should be adjusted too.

Maybe we should deprecate this function in favor of instance methods (e.g. mpfr.digits())?  I think that a generic function, returning a string - is better.  Something like ``to_str(num, base=10, prec=0, /)``, which will be a wrapper to ``mpfr_out_str()`` for the mpfr type.

skirpichev:
0ce584dd Correct signature for gmpy2.digits(), closes #504
aea8a0f9 Fix -Wunused-but-set-variable (drop unused copy variable)
3e013e7c Gitignore per-project directories for virtual environments
22763582 Drop support for the CPython 3.7
29aebd3c Don't upload code coverage to Codecov

### DIFF
--- a/.github/workflows/pip_install_gmpy2.yml
+++ b/.github/workflows/pip_install_gmpy2.yml
@@ -63,14 +63,6 @@ jobs:
           PYTHONPATH=`pwd`/gmpy2 python test_cython/runtests.py
           lcov --capture --directory . --no-external --output-file build/coverage.info
           genhtml build/coverage.info --output-directory build/coverage
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v4
-        with:
-          gcov: true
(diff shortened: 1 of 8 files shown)


casevh:
Thanks.

PR #525, opened 2024-10-17
Avoid OverflowError in mpz.__rshift__()

### LINKED ISSUE #524: mpz(1)>>111111111111111111111 raises OverflowError
Probably, we should just return ``0`` instead (if the rhs doesn't fit in ``mp_bitcnt_t``), something that CPython does.

Perhaps, this will address that Sage failure: https://github.com/sagemath/sage/pull/38565#issuecomment-2311204724

I'll work on a patch.

skirpichev:
Closes #524

### DIFF
--- a/src/gmpy2_mpz_bitops.c
+++ b/src/gmpy2_mpz_bitops.c
@@ -592,8 +592,35 @@ GMPy_MPZ_Rshift_Slot(PyObject *self, PyObject *other)
     MPZ_Object *result, *tempx;
 
     count = GMPy_Integer_AsMpBitCnt(other);
-    if ((count == (mp_bitcnt_t)(-1)) && PyErr_Occurred())
-        return NULL;
+    if ((count == (mp_bitcnt_t)(-1)) && PyErr_Occurred()) {
+        if (!PyErr_ExceptionMatches(PyExc_OverflowError)) {
+            /* LCOV_EXCL_START */
+            return NULL;
+            /* LCOV_EXCL_STOP */
+        }
+        PyErr_Clear();
+
+        PyObject *tmp = PyNumber_Long(other);
+
+        if (!tmp) {
+            /* LCOV_EXCL_START */
+            return NULL;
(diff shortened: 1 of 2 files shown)


skirpichev:
@casevh, second patch removes scripts/mp_bitcnt_t.diff, thus I'm marking this as a draft.

It's possible to implement this patch in other way, but I think it triggers a bug in the patch.  I think that proper support for 64-bit mp_bitcnt_t on win64 1) requires a more work, 2) we should submit patch to the upstream first.

casevh:
What bug is triggered in the mp_bitcnt_t patch? A previous bug that I reported upstream was fixed within a few days.

skirpichev:
> What bug is triggered in the mp_bitcnt_t patch?

See failed job: https://github.com/aleaxit/gmpy/actions/runs/11379818079/job/31657941881

It's on Windows, I can't reproduce this.  I guess, mpz_fdiv_q_2exp() make some assumptions on max value of the last argument.

skirpichev:
@casevh, I think failure with ``mp_bitcnt_t.diff`` happens, because ``mpz_fdiv_q_2exp()`` uses ``mpn_rshift()`` to do divisions.  This will not work with fourth argument of mp_bitcnt_t type, because it's declared as
```c
mp_limb_t mpn_rshift (mp_ptr, mp_srcptr, mp_size_t, unsigned int);
```
I don't see an easy fix.

What do you think?

casevh:
Per the GMP documentation, (the fourth argument) must be in the range 1 to mp_bits_per_limb{}-1. So I think there is a bug somewhere else. I'll try to step through the logic in cfdiv_q_2exp.

skirpichev:
FYI, failed test: ``assert (mpz(123) >> 111111111111111111111) == mpz(0)``

skirpichev:
@casevh, you were right - this was my fault.  Now patch should be fixed.

casevh:
Thanks.

Issue #522, opened 2024-10-14
Still having large bit precision calculation issues


wjamesjr:
When using mpfrs at high precision (> 2 ** 30 bits) algorithms fail.  I have verified that mpfr works as expected using a C++ wrapper and test pi agm code.  Using very similar code with gmpy, the code fails (agm fails to converge).  I am attaching some test code to demonstrate the problem.  At 250 million digits the agm routine works, at 400 million digits the agm routine fails to converge.  The c++ wrapper had a similar issue that was determined to be an argument in a call that needed to be a long that was an int.
[PiAGM_GMPY.py.zip](https://github.com/user-attachments/files/17367191/PiAGM_GMPY.py.zip)

skirpichev:
> Using very similar code with gmpy

Attached archive has only Python code.  We can't guess how "similar" it is to your C/C++ code.

wjamesjr:
C++ code is attached.
[main.cpp.zip](https://github.com/user-attachments/files/17389675/main.cpp.zip)

skirpichev:
Well, maybe I miss something else (versions use different variable name conventions, etc; it's not trivial to compare literally), but so far you code examples aren't "identical".  For example:
* to set precision you are using ``math.log(10) / math.log(2)`` coefficient, instead of more precise ``math.log2(10)``.  Either way it's not an equivalent of ``mpfr::digits2bits()``.
* you are using different initial values for ``y0``/``g``.  ``sqrt(0.5) != 1/sqrt(mpfr("2.0")``.
* finally, ``mpf("10.E-digits")`` != ``mpfr("10.0")**-digits``.

So, if you C++ code works correctly, please double check that Python code match it.

wjamesjr:
Made changes to Python source as suggested.  Still exhibits same issue with large bit precision.
[PiAGM_GMPY.py.zip](https://github.com/user-attachments/files/17396320/PiAGM_GMPY.py.zip)

skirpichev:
So far you just changed math.log(10) / math.log(2) to math.log2(10)

wjamesjr:
**bcf = math.log2(10)**

g, z, half = **sqrt(mpfr("0.5"))**, mpfr("0.25"), mpfr("0.5")

**epsilon = mpfr("10.E-" + str(needed_digits))**

The above three lines were changes.

casevh:
[...]

PR #526, opened 2024-10-23
Prevent crash in GMPy_MPFR_Format()

skirpichev:
Closes https://github.com/aleaxit/gmpy/issues/522

See https://github.com/aleaxit/gmpy/issues/522#issuecomment-2431541545


casevh:
> @casevh, should I rephrase error message? E.g. include references to digits methods/function?

Yes, I think that would be great. Something like "The maximum precision for string formatting exceeded. Please use gmpy2.digits() or mpft.digits()."

skirpichev:
@casevh, rephrased.

PR #528, opened 2024-11-13
Misc fixes

skirpichev:
* ~~Update mpz_set_PyLong() signature (return an integer)~~
* Bump cibuildwheel to v2.21.3
* Update src/pythoncapi_compat.h
* Use PyLong_IsNegative()

Note: first commit taken from #495.  I hope that API will be accepted, thus return value for mpz_set_PyLong() will be required.  I would prefer include this change here to cleanup #495.  @casevh, if you aren't ok with this - I'll remove this commit.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -23,7 +23,7 @@ jobs:
           path-type: inherit
         if: ${{ matrix.os == 'windows-2022' }}
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.18.0
+        uses: pypa/cibuildwheel@v2.21.3
         env:
           CIBW_PRERELEASE_PYTHONS: True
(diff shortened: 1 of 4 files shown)


casevh:
Is there any dependance on PEP 757? If so, should we wait for PEP 757 to be accepted?

skirpichev:
> Is there any dependance on PEP 757?

@casevh, first commit currently is useless.  On another hand, it does nothing bad; I think most compilers nowadays will optimize out added branches.

On pros: this will make https://github.com/aleaxit/gmpy/pull/495 smaller.

skirpichev:
BTW, different signature for mpz_set_PyLong() will be required also for PyPy support.

skirpichev:
Ok.  @casevh, first commit was reverted.

casevh:
Thanks.

PR #534, opened 2024-12-02
Fix mpz/mpq.__float__, add tests

### LINKED ISSUE #533: float(mpz(int)) != float(int)
An example (assuming CPython float's are IEEE 754 doubles):
```pycon
>>> x=225_188_150_488_381_457
>>> float(x)
2.2518815048838147e+17
>>> float(mpz(x))
2.2518815048838144e+17
```
I didn't test too much edge cases, but I would expect that overflow conditions also could be slightly different.

On another hand, conversion to mpfr (in appropriate context) and then to a float - works as expected:
```pycon
>>> set_context(ieee(64))
>>> float(mpfr(x))
2.2518815048838147e+17
>>> float.fromhex(f"{mpfr(x):a}")
2.2518815048838147e+17
```

skirpichev:
Closes #533

### DIFF
--- a/src/gmpy2_convert_gmp.c
+++ b/src/gmpy2_convert_gmp.c
@@ -159,23 +159,69 @@ GMPy_MPZ_Int_Slot(MPZ_Object *self)
     return GMPy_PyLong_From_MPZ(self, NULL);
 }
 
-static PyObject *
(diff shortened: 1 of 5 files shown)


casevh:
Thanks.

PR #535, opened 2024-12-09
Update mpz_set_PyLong() signature (return an integer)

skirpichev:
The PEP 757 is accepted.  This pr factor out required changes in the mpz_set_PyLong() signature to make #495.  Once merged, I'll rebase mentioned pr.


casevh:
Thanks.

PR #543, opened 2024-12-28
Drop support for CPython 3.8

skirpichev:


casevh:
Should we release gmpy2-2.2.2 before we drop support for CPython 3.8?

Are there any changes that should be done before a 2.2.2 release?

skirpichev:
You always can create a branch for 2.2.x to push more fixes to one.  I think that removal of 3.8 support in the master is irrelevant for this.  2.2.x should continue support 3.8.

skirpichev:
@casevh, does it make sense for you now?  Both mpmath and sympy dropped support for CPython 3.8.

casevh:
I'm okay with dropping support for 3.8.

casevh:
I'll re-run the SPDX tests.

PR #548, opened 2025-03-15
Create SECURITY.md

DavidOsipov:
I don't know active maintainer's email, so just placed a placeholder. 

This library is being used by guys like me for building cryptolibs and only recently I've read this:

> gmpy2 can crash the Python interpreter in case of memory allocation failure. To mitigate this feature of memory management in the GMP library, you should estimate the size of all results and prevent calculations that can exaust available memory.

If I haven't gone for info about how C part of gmpy2 was compiled, I would have missed this vulnerability. The gmpy2 should have a Security policy and a proper vulnerability reporting and disclosing mechanism to prevent this.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Supported Versions
+
+Below are the versions of gmpy2 currently supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 2.1.x   | :white_check_mark: |
+| 2.0.x   | :x:                |
+| < 2.0   | :x:                |
+
+## Reporting a Vulnerability
+
+To report a vulnerability, please email [maintainer-email@example.com] or open an issue on GitHub (preferably marked as private if sensitive). We’ll aim to respond within 5 business days. If the vulnerability is accepted, we’ll work on a fix and coordinate disclosure; if declined, we’ll provide reasoning. For urgent issues, please indicate priority in your report.


skirpichev:
I'm not sure that gmpy2 is suitable for applications, that expose it's functional to untrusted third parties.  This project rather provide a fast integer arithmetic or arbitrary precision floating-point numbers for scientific packages (sagemath, sympy, mpmath, etc).

Mentioned vulnerability is a "feature" of the GMP memory management (see chapter 13 of the GMP manual, Custom Allocation).  Hardly we can mitigate one in the gmpy2 project.

As a bit of advertising, you can try the [python-gmp](https://github.com/diofant/python-gmp), in case you need only integers/rationals and few common functions on integers.  It's on early stage and requires a huge refactoring, but some projects that depend on the gmpy2 (i.e. mpmath or diofant) can work on top of provided module (CI tests pass, etc).

DavidOsipov:
The trouble can be mitigated for my project. The main issue is that I didn't know about it on the firsthand. The won't fix vulnerabilities should still have appropriate CVEs for other dev to consider their usage and apply mitigations.

skirpichev:
@casevh, does it make sense for you in some form?

If so, probably you could set up private vulnerability reporting for this repository: https://docs.github.com/en/code-security/security-advisories/working-with-repository-security-advisories/configuring-private-vulnerability-reporting-for-a-repository

Then people could fill reports like this: https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability

skirpichev:
Sorry, thanks, but I don't think we need this.

PR #553, opened 2025-04-14
Correct optimization for negative ndigits in mpz.__round__()

### LINKED ISSUE #552: Incorrect optimization in mpz.__round__()
For negative number of digits we have following specialization:
https://github.com/aleaxit/gmpy/blob/ea6cdf16064e8f5bd575c0c1b02df0ea0fa70fc4/src/gmpy2_mpz_misc.c#L270-L272

This seems invalid, the reproducer (CPython lacks optimization for negative `round_digits`, see https://github.com/python/cpython/issues/132474):
```pycon
>>> import gmpy2
>>> gmpy2.__version__
'2.2.1'
>>> round(gmpy2.mpz(501), gmpy2.mpz(-3))
mpz(0)
>>> round(int(gmpy2.mpz(501)), int(gmpy2.mpz(-3)))
1000
```

skirpichev:
The GNU GMP manual says that about mpz_sizeinbaze(): "The result will be either exact or 1 too big."  First happened for example in the issue. We should use strict inequality to handle this.

Closes #552

### DIFF
--- a/src/gmpy2_mpz_misc.c
+++ b/src/gmpy2_mpz_misc.c
@@ -267,7 +267,7 @@ GMPy_MPZ_Method_Round(PyObject *self, PyObject *const *args,
     round_digits = -round_digits;
 
     if ((result = GMPy_MPZ_New(NULL))) {
-        if ((unsigned)round_digits >= mpz_sizeinbase(MPZ(self), 10)) {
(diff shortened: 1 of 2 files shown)


casevh:
I implemented the same fix locally and ran some brute-force tests. I'll commit your PR after the checks complete.

PR #571, opened 2025-06-18
Support mixed-mode arithmetic for mpc

### LINKED ISSUE #520: Support mixed-mode arithmetic rules for mpc & real types?
The MPC has special routines to do arithmetic with operands of different types (i.e. mpc_t and mpfr_t), but gmpy2 don't use this stuff (for example, mpc_mul_fr()).

I think we should.  Such arithmetic fix some analytical identities, that are broken in gmpy2 now.  Consider asinh:
```pycon
>>> import gmpy2
>>> z = gmpy2.mpc(-0.0, 2)
>>> gmpy2.asinh(z)
mpc('-1.3169578969248168+1.5707963267948966j')
>>> gmpy2.log(z + gmpy2.sqrt(1 + z*z))  # wrong real part, from other side of the branch cut
mpc('1.3169578969248166+1.5707963267948966j')
```
Also, I would expect some performance boost.

See also https://github.com/python/cpython/pull/124829

skirpichev:
Closes #520

### DIFF
--- a/src/gmpy2_add.c
+++ b/src/gmpy2_add.c
@@ -268,6 +268,46 @@ GMPy_Complex_AddWithType(PyObject *x, int xtype, PyObject *y, int ytype,
         return (PyObject*)result;
     }
 
+    if (IS_TYPE_COMPLEX(xtype) && IS_TYPE_REAL(ytype)) {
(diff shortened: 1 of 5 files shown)


casevh:
Thanks!

PR #574, opened 2025-06-20
Use context's rounding mode per default in format()

### LINKED ISSUE #503: format() should take into account rounding mode of the context?
Reproducer:
```pycon
>>> from gmpy2 import *
>>> format(mpfr(2.675), '.2f')
'2.67'
>>> get_context().round = RoundUp
>>> format(mpfr(2.675), '.2f')
'2.67'
```

I'm not sure whether it's a bug or a feature.

On one hand, changing rounding direction does affect C printf():
```
$ cat a.c
#include <fenv.h>
#include <stdio.h>
int main(void)
{
    double x = 2.675;
    fesetround(FE_TONEAREST);
    printf("%.2f\n", x);
    fesetround(FE_UPWARD);
    printf("%.2f\n", x);
    return 0;
}
$ gcc a.c -std=c11 -Wpedantic -lm && ./a.out
2.67
2.68
```
On another hand, current behaviour is consistent with CPython, which just does round to nearest (with ties to even), unless it's configured to use libc's printf (I'm not sure if this option is still available).

skirpichev:
Closes #503

### DIFF
--- a/src/gmpy2_format.c
+++ b/src/gmpy2_format.c
@@ -208,6 +208,11 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
     int buflen;
     int seensign = 0, seenalign = 0, seendecimal = 0, seendigits = 0;
     int seenround = 0, seenconv = 0;
+    CTXT_Object *context = NULL;
+    mpfr_rnd_t ctx_round;
+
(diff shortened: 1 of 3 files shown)


casevh:
Thanks.

PR #595, opened 2025-10-02
Update cibuildwheel confg

skirpichev:
This include https://github.com/aleaxit/gmpy/pull/594


casevh:
Thanks.

skirpichev:
Ah, my mistake.  The pyproject.toml's part was taken from the python-gmp, which miss such wheels, like the python-flint.

Though, I'm not sure if this should be reverted, especially i686.  There should be a tradeoff between providing wheels for less common systems and time, required for CI jobs.

mkoeppe:
> There should be a tradeoff between providing wheels for less common systems and time, required for CI jobs.

It can certainly make sense to restrict the platforms tested for CI jobs running on PRs.

But for releases, I'll say that it's the duty of packages as basic as `gmpy2` to strive for the most complete platform support.

FWIW, in passagemath I build wheels for musllinux but not for i686.

PR #606, opened 2025-11-16
Set Py_MOD_GIL_USED if the MPFR is not thread-safe

### LINKED ISSUE #599: Inspecting whether MPFR was built with `--enable-thread-safe`
Per the documentation:

>The gmpy2 has support for [:term:`free threading`](https://github.com/aleaxit/gmpy/blob/a3c7ca8c136a1ac25092756e45fedda61896de1d/docs/overview.rst#id1) builds of the CPython. But this requires using thread-local storage for the MPFR data, such as as flags, the exponent range, the default precision, etc. That is, the MPFR should be build with the --enable-thread-safe configure option.

Is there a way to inspect whether MPFR was built with this option at runtime within Python? This would be useful in a downstream test suite, where I don't want to attempt testing code that uses mpmath in parallel if that would result in failures in gmpy2.

It would appear that MPFR has the function `mpfr_buildopt_tls_p()`, which could be used to provide this information.

skirpichev:
Closes #599

### DIFF
--- a/src/gmpy2.c
+++ b/src/gmpy2.c
@@ -942,7 +942,11 @@ static PyModuleDef_Slot Pygmpy_slots[] = {
      Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED},
 #  endif
 #  if PY_VERSION_HEX >= 0x030D0000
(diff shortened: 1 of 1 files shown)


casevh:
Thanks for all your support.

PR #619, opened 2025-11-27
pyproject.toml: Allow setuptools >= 80, use modern editable for lcov

mkoeppe:
To my understanding, pinning setuptools < 80 is only needed for the `lcov` tests, for which the pinning is already attempted separately:
https://github.com/gmpy2/gmpy2/blob/649f2a63dba60cd5e6a38fce7de5b279c1895a40/.github/workflows/pip_install_gmpy2.yml#L48

(This attempt only fails because it ignores build isolation.)

We use a `constraints.txt` for this pin for the lcov test.

### DIFF
--- a/.github/workflows/pip_install_gmpy2.yml
+++ b/.github/workflows/pip_install_gmpy2.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


skirpichev:
Ok, I don't see any progress here.  And this PR has merge conflicts.  Lets close.

Please reopen if you are willing to continue this work.

PR #708, opened 2026-07-02
Support '=' alignment in mpfr formatting

### LINKED ISSUE #505: Invalid conversion specification (ValueError) for align '=' value in format spec
```pycon
>>> from gmpy2 import *
>>> f"{mpfr('123.456'):=.5g}"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid conversion specification
>>> f"{float(mpfr('123.456')):=.5g}"  # works for python floats
'123.46'
```

I think we should be compatible with python floats here.

apoorva-01:
`format(mpfr('123.456'), '=.5g')` raised `ValueError: Invalid conversion specification` where a Python float gives `'123.46'`. The `=` (sign-aware padding) alignment wasn't accepted. As you pointed out the same gap was in the integer types, so this now covers `mpz` and `xmpz` as well.

gmpy2 hands width and alignment to `str.__format__` on the already-formatted number, and `str` can't do `=` (it's numeric-only). The fix parses `=` and pads in C, putting the fill between the sign (and any `0x`/`0o`/`0b` prefix) and the digits, so it matches CPython for both the mpfr and the mpz/xmpz paths. mpc is left as-is since its docstring already documents `=` and zero-padding as unsupported.

The one I left out is `mpq`: it has no `__format__` at all right now (any non-empty spec raises TypeError), so `=` there really means giving mpq format support from scratch, which feels like its own change rather than part of this fix. Happy to do that separately if you'd like. Tests are in `test_mpz_format` and `test_mpfr_format` and fail without the change; they check both signs, zero-fill, the `#` base prefix, and inf/nan against CPython's output.

Closes #505

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -7,7 +7,6 @@ Changes in gmpy2 2.4.0
 ----------------------
 
 * Drop support for CPython < 3.11. (skirpichev)
-* Support '=' alignment option when formatting mpz, xmpz and mpfr. (apoorva-01)
 
 Changes in gmpy2 2.3.0
 ----------------------
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -7,7 +7,7 @@ Changes in gmpy2 2.4.0
 ----------------------
 
 * Drop support for CPython < 3.11. (skirpichev)
-* Support '=' alignment option in floating-point formatting. (apoorva-01)
+* Support '=' alignment option when formatting mpz, xmpz and mpfr. (apoorva-01)
 
 Changes in gmpy2 2.3.0
 ----------------------
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -7,6 +7,7 @@ Changes in gmpy2 2.4.0
 ----------------------
 
 * Drop support for CPython < 3.11. (skirpichev)
+* Support '=' alignment option in floating-point formatting. (apoorva-01)
 
 Changes in gmpy2 2.3.0
 ----------------------
--- a/src/gmpy2_format.c
+++ b/src/gmpy2_format.c
@@ -180,6 +180,38 @@ GMPy_MPZ_Format(PyObject *self, PyObject *args)
         } \
     } while (0);
 
+/* Sign-aware padding for the '=' alignment option. The numeric string 's'
+ * already carries its sign (if any); the fill is inserted between that sign
+ * and the digits until the requested field width is reached. This mirrors
+ * Python's '=' alignment, which str.__format__() can't do for us. */
+static PyObject *
+GMPy_MPFR_Pad_Equal(const char *s, long width, char fill)
+{
+    size_t slen = strlen(s);
+
+    if (width <= 0 || (size_t)width <= slen) {
+        return PyUnicode_FromString(s);
+    }
+
+    size_t pad = (size_t)width - slen;
+    int has_sign = (s[0] == '+' || s[0] == '-' || s[0] == ' ');
+    char *buf = malloc((size_t)width + 1), *q;
+    PyObject *result;
+
+    if (!buf) {
+        return PyErr_NoMemory();
+    }
+    q = buf;
+    if (has_sign) {
+        *(q++) = s[0];
+    }
+    memset(q, fill, pad);
+    strcpy(q + pad, s + has_sign);
+    result = PyUnicode_FromString(buf);
+    free(buf);
+    return result;
+}
+
 PyDoc_STRVAR(GMPy_doc_mpfr_format,
 "__format__($self, fmt, /)\n--\n\n"
 "Return a Python string by formatting self using the format string 'fmt'.\n\n"
@@ -209,7 +241,9 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
     char mpfrfmt[100], fmt[30];
     int buflen;
     int seensign = 0, seenalign = 0, seendecimal = 0, seendigits = 0;
-    int seenround = 0, seenconv = 0, seenindicator = 0;
+    int seenround = 0, seenconv = 0, seenindicator = 0, seenequal = 0;
+    long equalwidth = 0;
+    char equalfill = ' ';
     CTXT_Object *context = NULL;
     mpfr_rnd_t ctx_round;
 
@@ -244,6 +278,17 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
             else {
                 *(p3++) = *p1;
                 seenalign = 1;
+                continue;
+            }
+        }
+        if (*p1 == '=') {
+            if (seenalign || seensign || seendecimal || seendigits || seenround || seenindicator) {
+                VALUE_ERROR("Invalid conversion specification");
+                return NULL;
+            }
+            else {
+                seenalign = 1;
+                seenequal = 1;
                 continue;
             }
         }
@@ -297,6 +342,20 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
             }
             else if (seendecimal) {
                 *(p2++) = *p1;
+                continue;
+            }
+            else if (seenequal) {
+                /* A leading '0' selects zero fill; the rest is the width. */
+                if (equalwidth == 0 && equalfill == ' ' && *p1 == '0') {
+                    equalfill = '0';
+                }
+                else {
+                    equalwidth = equalwidth * 10 + (*p1 - '0');
+                    if (equalwidth > 1000000) {
+                        VALUE_ERROR("too long format string");
+                        return NULL;
+                    }
+                }
                 continue;
             }
             else {
@@ -429,11 +488,13 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
         strcat(newbuf, buffer);
         strcat(newbuf, ".0");
         mpfr_free_str(buffer);
-        mpfrstr = PyUnicode_FromString(newbuf);
+        mpfrstr = seenequal ? GMPy_MPFR_Pad_Equal(newbuf, equalwidth, equalfill)
+                            : PyUnicode_FromString(newbuf);
         free(newbuf);
     }
     else {
-        mpfrstr = PyUnicode_FromString(buffer);
+        mpfrstr = seenequal ? GMPy_MPFR_Pad_Equal(buffer, equalwidth, equalfill)
+                            : PyUnicode_FromString(buffer);
         mpfr_free_str(buffer);
     }
     if (!mpfrstr) {
--- a/src/gmpy2_format.c
+++ b/src/gmpy2_format.c
@@ -31,6 +31,43 @@ PyDoc_STRVAR(GMPy_doc_mpz_format,
 "format types, `mpfr` is used to convert the integer to a floating-point\n"
 "number before formatting.");
 
+/* Sign-aware padding for the '=' alignment option. The numeric string 's'
+ * already carries its sign and any base prefix (0x/0o/0b); the fill is inserted
+ * between that head and the digits until the requested field width is reached.
+ * This mirrors Python's '=' alignment, which str.__format__() can't do for us. */
+static PyObject *
+GMPy_Format_Pad_Equal(const char *s, long width, char fill)
+{
+    size_t slen = strlen(s);
+
+    if (width <= 0 || (size_t)width <= slen) {
+        return PyUnicode_FromString(s);
+    }
+
+    size_t pad = (size_t)width - slen, head = 0;
+    char *buf;
+    PyObject *result;
+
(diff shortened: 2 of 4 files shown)


docs/history.rst
```diff
@@ -7,6 +7,7 @@ Changes in gmpy2 2.4.0
 ----------------------
 
 * Drop support for CPython < 3.11. (skirpichev)
+* Support '=' alignment option in floating-point formatting. (apoorva-01)
```
skirpichev: That's fine, but next time please don't update change log.  I prefer to do this separately, to reduce merge conflicts.

src/gmpy2_format.c
```diff
@@ -299,6 +344,20 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
                 *(p2++) = *p1;
                 continue;
             }
+            else if (seenequal) {
+                /* A leading '0' selects zero fill; the rest is the width. */
```
skirpichev, line 401: Please note that from [docs](https://docs.python.org/3.14/library/string.html#format-specification-mini-language):
> When no explicit alignment is given, preceding the width field by a zero ('0') character enables sign-aware zero-padding for numeric types, excluding [complex](https://docs.python.org/3.14/library/functions.html#complex).

This is not the case for your patch:
```pycon
>>> f"{mpfr(-1.1):020e}"
'0000000-1.100000e+00'
>>> f"{float(mpfr(-1.1)):020e}"
'-00000001.100000e+00'
```

src/gmpy2_format.c
```diff
@@ -299,6 +344,20 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
...
+                if (equalwidth == 0 && equalfill == ' ' && *p1 == '0') {
+                    equalfill = '0';
+                }
+                else {
+                    equalwidth = equalwidth * 10 + (*p1 - '0');
+                    if (equalwidth > 1000000) {
```
skirpichev: Looks as a magic constant for me.  Why is this?

src/gmpy2_format.c
```diff
@@ -429,11 +488,13 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
         strcat(newbuf, buffer);
         strcat(newbuf, ".0");
         mpfr_free_str(buffer);
-        mpfrstr = PyUnicode_FromString(newbuf);
+        mpfrstr = seenequal ? GMPy_MPFR_Pad_Equal(newbuf, equalwidth, equalfill)
```
skirpichev: I think this branch isn't tested by added tests, no?

test/test_mpfr.py
```diff
@@ -348,6 +348,18 @@ def test_mpfr_format():
...
+    assert f"{mpfr('123.456'):=10.2f}" == '    123.46'
+    assert f"{mpfr('-123.456'):=10.2f}" == '-   123.46'
+    assert f"{mpfr('123.456'):=+10.2f}" == '+   123.46'
+    assert f"{mpfr('123.456'):=010.2f}" == '0000123.46'
+    assert f"{mpfr('-123.456'):=010.2f}" == '-000123.46'
+    assert f"{mpfr('-inf'):=10.2f}" == '-      inf'
+    assert f"{mpfr('nan'):=10.2f}" == '       nan'
+    assert f"{mpfr('123.456'):= 10.2f}" == '    123.46'
+    pytest.raises(ValueError, lambda: format(mpfr(1), '*=10.2f'))
+
```
skirpichev: This can be tested differently, if you are just checking conformance wrt Python floats:
1. create a set of values to test
2. set of format string
3. generate and compare gmpy2 format wrt Python's for same value

See also mpmath/tests/test_format.py on how this can be tested with hypothesis.

skirpichev (changes requested):
Please note, that #505 is not restricted to the mpfr, it's valid for other gmpy2 types too (except for mpc).  Either remove from description claim that this pr fixed issue, or address it completely.

docs/history.rst
apoorva-01: Sorry about that, removed it. I'll leave the changelog to you.

src/gmpy2_format.c
apoorva-01: It's fixed. A leading `0` with no explicit alignment now goes through the same sign-aware path as `=` (both mpz and mpfr), so `f"{mpfr(-1.1):020e}"` gives `-00000001.100000e+00` now. Explicit `<`/`>`/`^` stay as they were.

src/gmpy2_format.c
apoorva-01: Replaced it with the overflow check from your `get_integer` in python-gmp (`equalwidth > (PY_SSIZE_T_MAX - digit)/10`) and made `equalwidth` a `Py_ssize_t` - so there is no arbitrary limit anymore

src/gmpy2_format.c
apoorva-01: Right, it wasn't. Added a case that hits it: `f"{mpfr('123'):=08.0f}"`, where the whole-number output takes the `.0`-append path and then gets padded.

test/test_mpfr.py
apoorva-01: Reworked both tests to compare gmpy2 against Python `float`/`int` over a set of values crossed with a set of format strings, like you suggested. I kept it a curated cross-product rather than hypothesis on purpose, so I could steer around two divergences that aren't this issue: gmpy2 appends `.0` to whole-number float output, and there's a separate pre-existing bug where `e`/`E` with an explicit precision builds a bad format string (`f"{mpfr('1.1'):.3e}"` gives `'%.3.6RNe'`). Happy to add a hypothesis pass on top if you'd like, and I can file that `e`-precision one separately.

apoorva-01:
Pushed a commit addressing the review. Summary of where the types stand:

- `mpz`/`xmpz` and `mpfr` now handle `=` and sign-aware zero-padding (including the implicit `0` case you pointed out).
- `mpc` I left out, since its docstring already documents `=` and zero-padding as unsupported.
- `mpq` has no `__format__` at all right now (any non-empty spec raises TypeError), so making `=` work there really means adding mpq formatting from scratch. That feels like its own change rather than part of this fix, but happy to do it as a follow-up if you'd like it in scope.

src/gmpy2_format.c
```diff
@@ -299,6 +397,28 @@ GMPy_MPFR_Format(PyObject *self, PyObject *args)
                 *(p2++) = *p1;
                 continue;
             }
+            else if (seenequal) {
+                /* A leading '0' selects zero fill; the rest is the width. */
+                if (equalwidth == 0 && equalfill == ' ' && *p1 == '0') {
+                    equalfill = '0';
+                }
+                else if (equalwidth > (PY_SSIZE_T_MAX - (*p1 - '0')) / 10) {
```
skirpichev, line 405: This still an arbitrary constraint.  Are you sure it's a possible case?

test/test_mpfr.py
```diff
@@ -348,6 +348,24 @@ def test_mpfr_format():
...
+    values = ['1.1', '-1.1', '123.456', '-123.456', '0.5', '-0.5']
+    specs = ['=10.2f', '=+10.2f', '= 10.2f', '=010.2f', '=.5g', '.5g',
+             '020e', '=020e', '>020e', '<020e', '^020e', '015.4f', '+015.4f',
+             '010.2f', '020.3f', '=020.3f', '020g', '=.2f']
+    for value in values:
+        for spec in specs:
+            assert format(mpfr(value), spec) == format(float(value), spec), (value, spec)
+    assert f"{mpfr(-1.1):020e}" == '-00000001.100000e+00'
+    assert f"{mpfr('-inf'):=10.2f}" == '-      inf'
+    assert f"{mpfr('nan'):=10.2f}" == '       nan'
```
skirpichev: Why you can't test this above?

test/test_mpfr.py
```diff
@@ -348,6 +348,24 @@ def test_mpfr_format():
...
+    # whole-number output goes through the '.0' path, then pads sign-aware
+    assert f"{mpfr('123'):=08.0f}" == '000123.0'
+    assert f"{mpfr('-123'):08.0f}" == '-00123.0'
```
skirpichev: Hmm, this is incompatible wrt the Python.  It's now a bug: https://github.com/gmpy2/gmpy2/issues/709.  Lets fix this in a separate pr.

test/test_mpfr.py
```diff
@@ -348,6 +348,24 @@ def test_mpfr_format():
...
+    pytest.raises(ValueError, lambda: format(mpfr(1), '*=10.2f'))
```
skirpichev: This is invalid:
```pycon
>>> format(float(1), '*=10.2f')
'******1.00'
>>> format(float(-1), '*=10.2f')
'-*****1.00'
```

test/test_mpz.py
```diff
@@ -659,6 +659,20 @@ def test_mpz_format():
...
+    values = [255, -255, 0, 7, -7, 1, -1]
+    specs = ['=10d', '=+10d', '= 10d', '=010d', '010d', '+010d', '08d',
+             '=#010x', '#010x', '010x', '=#010o', '#010o', '010o',
+             '=#010b', '#010b', '=#012o', '>08d', '<08d', '^08d']
+    for value in values:
+        for spec in specs:
+            assert format(mpz(value), spec) == format(value, spec), (value, spec)
+            assert format(xmpz(value), spec) == format(value, spec), (value, spec)
+    assert f"{mpz(255):=10.2f}" == '    255.00'  # float spec still delegates to mpfr
+    raises(ValueError, lambda: format(mpz(1), '*=10d'))
```
skirpichev: Again, this is invalid.

test/test_mpfr.py
```diff
@@ -359,6 +377,7 @@ def test_mpfr_format():
     assert f'{r:.2f}' == '2.68'
 
     pytest.raises(ValueError, lambda: format(mpfr(1), "1" + "0"*70))
+    pytest.raises(ValueError, lambda: format(mpfr(1), "=1" + "0"*70))
```
skirpichev: Are you sure this test adds anything?

skirpichev (commented):
> mpq has no `__format__` at all right now (any non-empty spec raises TypeError)

There is a dedicated issue for this.

src/gmpy2_format.c
apoorva-01: It's an overflow guard, not really a limit. A width with 19+ digits overflows `Py_ssize_t`, and this makes gmpy2 raise `ValueError` there the same way CPython does (`format(1, '9'*19 + 'd')` raises in both). 
Actually it's the same check as `get_integer` in python-gmp. happy to drop it if you want but then that input is signed overflow UB

test/test_mpfr.py
apoorva-01: Got it- it was redundant. Folded that case (and the inf/nan ones) into the conformance loop above and dropped the standalone asserts

test/test_mpfr.py
apoorva-01: Dropped those asserts and I'll leave that path to the separate PR.

test/test_mpfr.py
apoorva-01: Removed it.

test/test_mpz.py
apoorva-01: Removed.

test/test_mpfr.py
apoorva-01: Removed it. The `'1' + '0'*70` case just above already covers a too-long width.

test/test_mpfr.py
skirpichev: But exception is still present:
```pycon
>>> format(mpfr(1), '*=10.2f')
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    format(mpfr(1), '*=10.2f')
    ~~~~~~^^^^^^^^^^^^^^^^^^^^
ValueError: Invalid conversion specification
```

That should be fixed.

test/test_mpz.py
skirpichev: It should be fixed, not just removed.

test/test_mpfr.py
skirpichev: No, I don't think so: only exception message is same.  Please restore this test.

skirpichev (changes requested):
Please fix mentioned bugs (e.g. ``format(mpfr(1), '*=10.2f')``), fix broken CI.  Please restore tests, that cover overflow branches.

skirpichev:
@apoorva-01, let us know if you would like to finish this pr.

skirpichev:
Ok, closing.  Feel free to reopen if you address review.

